PR #58, opened 2026-04-24
fix(perf): skip (not fail) when baseline or current scene fails to load

RaananW:
## Summary

Perf tests in CI sometimes fail for reasons that are not actual performance regressions — most commonly when a PR adds new scenes whose baseline bundles don't yet exist, or when the baseline / current scene fails to load/render in CI. Today those throw `[NOT A PERFORMANCE ISSUE]` errors and red the entire perf suite.

This PR converts those cases into **skips with warning annotations**, and teaches the test-results reporter to surface them as pipeline warnings.

## Changes

- `tests/perf/perf-regression.spec.ts`
  - Baseline bundle missing → `test.skip()` + warning annotation (was already a skip, now annotated)
  - Baseline fails to load/render → `test.skip()` + warning (was: throw)
  - Current fails to load/render → `test.skip()` + warning (was: throw)
- `scripts/report-test-results.ts`
  - Parses `<skipped>` testcases and extracts Playwright `warning` annotations from the `<properties>` block
  - Emits `##vso[task.logissue type=warning]` for each
  - Marks the step `SucceededWithIssues` (yellow) when there are warnings but no true failures
  - Summary line now includes skipped count

## Effect

A PR that adds new scenes — or any environment where the baseline fails to render — will see the perf suite step turn **yellow** with clear warning annotations, instead of red. Genuine performance regressions still fail as before.


RaananW:
Converted to draft to test. i just added a temporary scene (which makes no sense, of course and will be removed). I just want to see the no-fail procedure (there should be a warning in the logs)

RaananW:
> Converted to draft to test. i just added a temporary scene (which makes no sense, of course and will be removed). I just want to see the no-fail procedure (there should be a warning in the logs)

ok, that worked :)

PR #62, opened 2026-04-28
dist package.json and rolled up d.ts

ryantrem:
This PR mainly adds two things:
1. A rolled up index.d.ts (alongside the rolled up index.js)
2. A generated package.json in the dist folder (with package name @babylonjs/lite)

This is enough to allow an `npm install <path to lite dist>` from the BJS repo for local tools development (it can be checked into BJS as a regular npm dependency once we are publishing the lite package).

Also included one fix to switch from `EngineContextInternal` to `EngineContext` in one API to keep that internal contract internal. Otherwise, we produce an index.d.ts with a reference to an unknown `EngineContextInternal`.

### DIFF
--- a/packages/babylon-lite/package.json
+++ b/packages/babylon-lite/package.json
@@ -28,6 +28,9 @@
         "dev": "vite build --watch"
     },
     "sideEffects": false,
+    "publishConfig": {
+        "types": "./dist/index.d.ts"
+    },
     "devDependencies": {
         "@webgpu/types": "^0.1.52",
         "typescript": "^5.7.0",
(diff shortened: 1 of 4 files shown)


deltakosh (approved):

RaananW:
I'll be copilot for a second:

You said the package name is "@babylonjs/lite", but it is called babylonjs-lite. was that intentional?

ryantrem:
> I'll be copilot for a second:
> 
> You said the package name is "@babylonjs/lite", but it is called babylonjs-lite. was that intentional?

The internal name (checked in src/package.json) is babylon-lite, but the public name (generated dist/package.json) is @babylonjs/lite: https://github.com/BabylonJS/Babylon-Lite/pull/62/changes?w=1#diff-996df714128ac9d5afb26658668005f81eb3c0aea2c9b0e1639a292aec889062R12

PR #68, opened 2026-05-01
Add support for EffectRenderer

deltakosh:
## Summary

EffectRenderer, and material path updates

## Validation
- pnpm test


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260501.11/testResults/parity/index.html)

<sub>Build 20260501.11 · merge @ 04e657b3f7e3c38b6c774fccb3e9a8274fcc3ea8</sub>

packages/babylon-lite/src/effect/effect-renderer.ts
```diff
@@ -0,0 +1,542 @@
...
+            pass.setPipeline(pipeline);
+            if (bindGroup) {
+                pass.setBindGroup(0, bindGroup);
+            }
+            pass.draw(3);
+            pass.end();
+            return 1;
+        },
+        dispose(): void {
+            task._renderPassDescriptor = { colorAttachments: [] };
```
Popov72, line 226: This task owns/builds `task._rt`, but `dispose()` only clears the render pass descriptor. When the scene frame graph disposes this task, the effect RTT color/depth textures remain alive. Please mirror `RenderPassTask.dispose()` and call `disposeRenderTarget(task._rt)` here before clearing descriptor state.

packages/babylon-lite/src/effect/effect-renderer.ts
```diff
@@ -0,0 +1,542 @@
...
+        record(): void {
+            buildRenderTarget(rt, eng);
+            patchColorAttachment(task, eng);
+        },
+        execute(): number {
+            const encoder = eng._currentEncoder;
+            if (!encoder) {
+                return 0;
+            }
+            ensureTargetSize(task, eng);
```
Popov72: This per-frame size check should not be needed here. Canvas-size render targets are rebuilt through the framework resize path (`_resize()` -> frame graph build -> `Task.record()`), and fixed RTTs do not change size during task execution. Keeping this in `execute()` duplicates render-target lifecycle work on the hot path.

packages/babylon-lite/src/effect/effect-renderer.ts
```diff
@@ -0,0 +1,542 @@
...
+            patchColorAttachment(task, eng);
+            const pipeline = getEffectPipeline(config.effect as EffectWrapperInternal, task._targetSignature);
```
Popov72: The effect pipeline and bind group should be resolved during `record()` and cached on the task, not fetched every frame in `execute()`. This task's render-target signature is fixed for its lifetime, so `execute()` should only patch per-frame attachment state, bind cached GPU state, and draw.

packages/babylon-lite/src/effect/effect-renderer.ts
```diff
@@ -0,0 +1,542 @@
...
+            if (er._disposed) {
+                return 0;
+            }
+            ensureRtCanvasSize(er._rt, er._engine);
+            applyColorAttachmentState(er._colorAttachment, er._rt, er._engine, er._clear, er.clearColor);
+            const encoder = er._engine._currentEncoder;
+            if (!encoder) {
+                return 0;
+            }
+            const pipeline = getEffectPipeline(er._effect, er._targetSignature);
```
Popov72: Same lifecycle concern here: the direct `EffectRenderer` target signature is fixed for the renderer lifetime, so the pipeline and bind group should be cached before the per-frame `_record()` path instead of being resolved every frame. The pipeline map can still stay on `EffectWrapper` for reuse across multiple RT signatures.

packages/babylon-lite/src/effect/effect-renderer.ts
```diff
@@ -0,0 +1,542 @@
...
+    if (typeof bindingNameOrIndex === "number") {
+        return layout.binding === bindingNameOrIndex;
+    }
+    return layout.name === bindingNameOrIndex || String(layout.binding) === bindingNameOrIndex;
+}
+
+function writeUniformSlot(wrapper: EffectWrapperInternal, slot: EffectUniformSlot, data: ArrayBuffer | ArrayBufferView): void {
+    const bytes = toBytes(data);
+    if (bytes.byteLength > slot.byteLength) {
+        throw new Error(`setEffectUniforms: ${bytes.byteLength} bytes exceeds uniform binding ${slot.layout.binding} size ${slot.byteLength}.`);
```
Popov72: Small diagnostic issue: this helper is `writeUniformSlot`, but the error message says `setEffectUniforms`. Please use the helper name here so failures point at the code path that actually detected the invalid byte length.

bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260504.1/testResults/parity/index.html)

<sub>Build 20260504.1 · merge @ 973ea7d85505d52ef3b2d32c17003ca308fff562</sub>

PR #70, opened 2026-05-04
Shadow Generator: skeletal animation + shadow-only PBR mode

ryantrem:
# Shadow Generator: skeletal animation + shadow-only PBR mode

This branch lands four related shadow + PBR features, plus the test/lab
coverage that exercises them. The original review surfaced a set of issues
(scene-id collisions, bundle-size regressions on PBR scenes, a couple of
guideline violations); the [Review-feedback follow-up](#review-feedback-follow-up)
section at the bottom summarises how each was addressed.

### 1. PBR `mode: "shadowOnly"` (commit `4e46c30`)

A new shading mode for `PbrMaterialProps` that mirrors BJS
`BackgroundMaterial.shadowOnly`: the surface is invisible except where a
shadow falls on it, in a caller-chosen colour. Useful for drop shadows
on a transparent ground plane, AR placement reticles, etc.

```ts
const ground = createGround(engine, { width: 30, height: 30 });
ground.material = createPbrMaterial({
    mode: "shadowOnly",
    color: [0, 0, 0],
    falloff: 1.0,
    alphaBlend: true,
});
ground.receiveShadows = true;
```

**Why a PBR-extension fragment, not a brand-new material type?** The
shadow-only output has to live on the same shading stage that already
knows how to consume the multi-light shadow buffer. PBR is where Lite's
multi-light + per-light shadow plumbing lives (`pbr-renderable.ts`,
`pbr-shadow-fragment.ts`, the multi-light loop's `shadowFactors`
array). Building a parallel material would mean re-implementing all of
that wiring; instead, `shadow-only-fragment.ts` slots into the existing
PBR `BC` template position and reuses everything below it. Cost when
unused: zero — the fragment, its UBO fields, and its WGSL only register
when at least one mesh declares `mode: "shadowOnly"`. The same fragment
also zeroes the `environmentIntensity`/`directIntensity` slots in its
`writeUbo` so the static PBR core stays free of mode-specific branches.

### 2. Discriminated union for `PbrMaterialProps` (commit `8dee3e2`)

`PbrMaterialProps` is now a discriminated union over a `mode` field:

- omit `mode` (or `"lit"`) → full PBR shading (default, same shape as
  before)
- `mode: "unlit"` → KHR_materials_unlit
- `mode: "shadowOnly"` → invisible except where shadow falls
- `mode: "skybox"` → sample IBL using the view direction

Each variant exposes only the props that are relevant to it. Common
flags (`alpha`, `alphaBlend`, `doubleSided`) live on a shared
`PbrMaterialPropsCommon` base. Authoring is now type-safe — picking
`mode: "shadowOnly"` only lets you set `color` and `falloff`, not
random PBR knobs that would have no effect.

Internally, `PbrMaterialPropsInternal` is the intersection of all
variants so the engine code paths can read any field without narrowing
on `mode` for every access.

### 3. Frustum-size override on the directional shadow generator (`81dca56`)

```ts
light.shadowGenerator = await createShadowGenerator(engine, light, casters, {
    frustumSize: 8, // half-extent in light space
    // ...
});
```

Mirrors BJS `DirectionalLight.shadowFrustumSize` paired with
`autoUpdateExtends = false`. When set, X/Y bounds are pinned to ±size
centered on the casters' AABB midpoint instead of being auto-fit. The
existing ESM blur can then spread a small caster's silhouette into a
soft drop shadow over the full frustum area — the natural pairing for
the new shadow-only ground.

### 4. Skinned shadow casting (`81dca56`)

The directional shadow generator now sorts casters into static vs
skinned. Skinned casters get a depth pipeline whose vertex shader
samples bone matrices from `skeleton.boneTexture`, giving the correct
silhouette for animated meshes. Two new vertex shaders cover the
4-bone and 8-bone cases (`shadow-skinned-{4,8}.vertex.wgsl`).

`renderShadowMap` always re-renders skinned casters (their bones can
change every frame even when the world matrix doesn't), while static
casters retain the existing world-matrix-version dirty tracking.

All skinned-shadow code lives in `shadow-skinned-caster.ts`, which is
**dynamic-imported only when the caster list contains a skinned mesh**.
Inside that module, the 4-bone and 8-bone vertex WGSL files are also
dynamic-imported via separate `import("…")` calls — a 4-bone-only scene
never fetches the 8-bone shader (and vice-versa). The shared BGL +
render pipeline are built once per skinning width per generator;
per-caster state is just a mesh UBO + bind group + cached world matrix.

### 5. `createShadowGenerator` is now async (`a69ff3a`)

```ts
light.shadowGenerator = await createShadowGenerator(engine, light, casters, cfg);
```

**Why async + dynamic import?** `GUIDANCE.md §4` is non-negotiable:
a scene that uses the shadow generator with only static casters must
not pay for the skinned-caster WGSL bytes. A static
`import "shadow-skinned-{4,8}.vertex.wgsl?raw"` at the top of
`shadow-generator.ts` would have added ~2.3 KB to every shadow-using
scene, busting the existing ceilings.

Solution: move all skinned-caster code into its own module
(`shadow-skinned-caster.ts`) and dynamic-`import` it from inside
`createShadowGenerator` only when at least one caster is skinned. The
module exposes a single `createSkinnedShadowHandle()` returning an
opaque `{ count, sync, draw }` handle — `shadow-generator.ts` never
sees any skinned-specific data structures, so its static bundle stays
lean.

That makes the function async — but every other "loads bytes to
construct" factory in the repo (`loadGltf`, `loadTexture2D`,
`createGroundFromHeightMap`, `loadDdsEnvironment`, …) is already async,
so the API shape is consistent with the rest of the codebase and
doesn't introduce a brand-new pattern. Static-only generators take the
fast no-await path inside the factory.

The `_pendingInit`-on-`ShadowGenerator` + `buildScene`-special-case
alternative was considered and rejected: it leaks shadow-generator
specifics into the generic boot path, and there's no precedent in the
repo for "sync factory that pushes deferred builders for its own
primary async work" — every factory using `_deferredBuilders` is also
already async.

## Tests + lab coverage (commits `8e7daeb`, `e8be097`, `2cb3147`)

- **`tests/unit/shadow-only-fragment.test.ts`** — 11 vitest cases for
  `shadowOnlyExt.detect`, `frag`, `writeShadowOnlyUBO` (defaults,
  no-op for non-shadowOnly modes, missing-offset safety), and the
  generated `BC` slot covering every `MAX_LIGHTS` index.

- **Scene 90 — PBR Shadow-Only Receiver** (procedural). Exercises
  `mode: "shadowOnly"` + `frustumSize` together. Lite uses a wide
  invisible PBR ground with `mode: "shadowOnly"`; the BJS oracle uses
  `BackgroundMaterial.shadowOnly` + `DirectionalLight.shadowFrustumSize`
  / `autoUpdateExtends = false`. Parity MAD 0.23.

- **Scene 91 — Skinned Shadow Casting** (Alien.gltf). Same skinned rig
  as scene 5, here registered as a shadow caster so the new skinned
  depth path is exercised end-to-end. Animation deterministic via
  `?seekTime=2`. Parity MAD 0.23. (The Alien glTF is a chest-up bust,
  not a full body, so the asset root is lifted by `+0.7` to sit on the
  ground rather than be buried.)

### `dot(finalSpecularScaled, …)` moved out of `pbr-template.ts` (`8e7daeb`)

The PBR template's alphaBlock unconditionally referenced
`finalSpecularScaled`, which is only declared by `ibl-fragment.ts`.
That meant any PBR material with `alphaBlend` but no IBL — including
shadow-only on a procedural scene — failed to compile. Fix: move the
line into the IBL fragment's `BA` slot alongside the existing
`finalRadianceScaled` reference, so it only emits when IBL is active.
Identical generated WGSL when IBL is on; clean compile when it's off.

## Review-feedback follow-up

Five issues raised on the original draft; all addressed in the merge
commit `cde25c6` and feedback commits `e0a382c` / `2cb3147`.

### 1. Scene-id collisions

Picked up after the merge with origin/master added scene 52 (RTT) /
74-76 (effect-renderer). Renumbered to **scene 90** (PBR Shadow-Only
Receiver) and **scene 91** (Skinned Shadow Casting) — `scene-config.json`
now reserves the 90+ block for this PR.

### 2. Bundle-size regressions on PBR scenes

Reviewer reported scene22 +5.2 KB, scene17 +3.0 KB, scene33 +2.6 KB,
scene67 +5.3 KB, "multiple PBR scenes around +1.9 KB". Fixes:

- Dropped the static `import { getDefaultBaseColorTexture, getDefaultOrmTexture }`
  from `pbr-pipeline.ts`. The bind-time path is back to master's
  `material.baseColorTexture!` / `material.ormTexture!`.
- Pre-population of the 1×1 fallback textures moved into a new
  `populatePbrDefaultTextures(engine, meshes)` helper inside
  `pbr-default-textures.ts`. The per-mesh loop now lives in the
  dynamic-imported chunk — static-only PBR scenes pay zero bytes for it.
- Consolidated the shadow-only fragment registration into the same
  `if (hasAnyShadowOnly || hasSkybox)` gate that triggers the defaults
  dynamic-import (one gate, was two).
- Moved the `material.mode === "shadowOnly"` zero-out for
  `environmentIntensity` / `directIntensity` from `writeMaterialData`
  into `shadowOnlyExt.writeUbo`. The static PBR core no longer branches
  on shadow-only.
- Restructured `shadow-skinned-caster.ts` to share one BGL + one
  pipeline per skinning width per generator (was per-caster), and to
  dynamic-import the 4-bone and 8-bone WGSL via separate
  `import("…")` calls.
- Consolidated `shadow-generator.ts`'s skinned support: a single
  `skinned: SkinnedShadowHandle | null` reference replaces the previous
  trio of helper-capture variables + per-caster array.

Result vs master baseline (post-merge):

| scene | master | now | ceiling | delta |
| --- | ---: | ---: | ---: | ---: |
| scene4 — shadows | 70.5 | 71.1 | 72.4 ✓ | +0.6 |
| scene8 — HDR glass | 73.6 | 74.0 | 74.2 ✓ | +0.4 |
| scene10 — pbr rough | 54.1 | 54.2 | 54.1 ✗ | +0.1 |
| scene12 — shader balls | 99.4 | 99.4 | 99.6 ✓ | 0 |
| scene17 — pbr-std thin | 85.1 | 85.4 | 89.8 ✓ | +0.3 |
| scene22 — pbr shadows | 94.9 | 95.7 | 101.8 ✓ | +0.8 |
| scene33 — lights punctual | 85.7 | 86.0 | 88.7 ✓ | +0.3 |
| scene67 — nme pbr core | 75.3 | 75.3 | 95 ✓ | 0 |
| scene90 — pbr shadow-only (new) | — | 71.8 | 80 ✓ | — |
| scene91 — skinned-shadow (new) | — | 104.9 | 130 ✓ | — |

Every reported regression is down to ≤+0.8 KB (was up to +5.3 KB) and
every scene is under its ceiling **except scene10**, which is +0.1 KB
over (54.2 vs 54.1). That residue is the discriminated-union
mode-checks (`mat.mode === "shadowOnly"` vs the old boolean flags) plus
the `hasAnyShadowOnly` local + dynamic-import gate, propagated to every
PBR scene. To clear scene10 without removing functionality I'd need a
small ceiling bump (54.1 → 54.3) — flagging here for explicit approval
per the immutable-ceiling rule.

### 3. Module-level `WeakMap` in `pbr-default-textures.ts`

Now lazy-initialized per GUIDANCE §4:
`let _baseColorCache: WeakMap<…> | null = null;` with `??=` inside the
getter. Combined with #2 above, the module is no longer in the static
graph at all unless a scene actually needs it.

### 4. Skinned-caster pipeline duplicated per mesh

Fixed: `shadow-skinned-caster.ts` now builds **one** BGL + **one**
render pipeline per (4-bone, 8-bone) variant per generator. Each caster
keeps only its mesh UBO, bind group, and cached world matrix. The draw
loop also avoids redundant `setPipeline` calls for consecutive casters
of the same width (which is the common all-4-bone case).

### 5. 8-bone shader fetched even when only 4-bone is needed

Fixed: the 4-bone and 8-bone vertex WGSL files are imported via
**separate** dynamic `import("…")` calls inside
`createSkinnedShadowHandle`, gated by `skinnedMeshes.some(...)` checks
on `joints1Buffer`. A 4-bone-only scene fetches only the 4-bone shader
(and vice-versa).

## Validation

- `pnpm lint` — clean
- `pnpm test:parity` — 118 passed, 1 failed (scene10 +0.1 KB; see above)
- `npx vitest run` — 104 passed (includes the 11 new shadow-only-fragment cases)
- Perf tests not run (per `AGENTS.md`, those are user/CI-only)


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.3/testResults/parity/index.html)

<sub>Build 20260505.3 · merge @ 22e1b8031b6d9625a724ac7ff96fe0de8f4d9cdb</sub>

bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.3/testResults/perf/index.html)

<sub>Build 20260505.3 · merge @ 22e1b8031b6d9625a724ac7ff96fe0de8f4d9cdb</sub>

deltakosh (approved):

bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.18/testResults/parity/index.html)

<sub>Build 20260505.18 · merge @ 9b71d08fe990f7d3f2481fbf3f2bdfeff7b29d5f</sub>

packages/babylon-lite/src/material/pbr/pbr-mesh-features.ts
```diff
@@ -39,7 +39,7 @@ export interface PbrFeatureCtx {
 
 /** Compute the `(features, features2)` bit pair for a single PBR mesh. */
 export function computeMeshPbrFeatures(mesh: Mesh, scene: SceneContext, ctx: PbrFeatureCtx): { features: number; features2: number } {
-    const mat = mesh.material as PbrMaterialProps;
+    const mat = mesh.material as PbrMaterialPropsInternal;
     const mi = mesh as import("../../mesh/mesh.js").MeshInternal;
```
Popov72, line 43: Now that this function can see `PbrMaterialPropsInternal`, `mode: "shadowOnly"` should force the alpha-blend feature in the nearby `hasAlphaBlend` decision. The public API docs say shadow-only is invisible except where a shadow falls, but if callers create `createPbrMaterial({ mode: "shadowOnly" })` without also setting `alphaBlend: true`, this keeps `PBR_HAS_ALPHA_BLEND` off, so the pipeline writes depth and returns an opaque black receiver instead of blending by the shadow alpha. Please include `mat.mode === "shadowOnly"` in that feature decision, or otherwise make the mode self-contained.

packages/babylon-lite/src/material/pbr/pbr-default-textures.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** Fill in `baseColorTexture` / `ormTexture` for every shadow-only / skybox mesh that
+ *  hasn't supplied its own. The loop lives here (in the dynamic-imported chunk) so the
+ *  static `pbr-renderable` bundle only contains the gated dynamic-import call. */
+export function populatePbrDefaultTextures(engine: EngineContext, meshes: readonly Mesh[]): void {
+    for (const m of meshes) {
+        const mat = m.material as PbrMaterialPropsInternal;
+        if (mat.mode === "shadowOnly" || mat.mode === "skybox") {
+            if (!mat.baseColorTexture) {
+                mat.baseColorTexture = getDefaultBaseColorTexture(engine);
```
Popov72, line 59: If we keep this default-texture approach, these cached fallback textures are later treated as normal material-bound textures: `collectPbrBoundTextures()` acquires them, mesh disposal releases them, and `releaseTexture()` destroys the underlying `GPUTexture` at refcount 0. The WeakMap still keeps the stale `Texture2D`, so a later shadow-only/skybox material on the same engine can bind a destroyed texture. That said, see my later comment on this module: I think the better fix is to avoid this default-texture module entirely by making the PBR template not declare/sample textures that a mode does not need.

packages/babylon-lite/src/material/pbr/fragments/shadow-only-fragment.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Mirrors BJS `BackgroundMaterial.shadowOnly`: the surface is invisible except
+ * where shadow falls on it. Lit fragments output alpha=0 (fully transparent);
+ * shadowed fragments output alpha proportional to the shadow strength, in a
+ * caller-chosen `shadowOnlyColor` (defaults to black).
+ *
+ * Only bundled when at least one mesh in the scene uses `mat.shadowOnly === true`.
+ *
+ * Implementation notes
+ * --------------------
+ * The PBR template's multi-light block declares `var shadowFactors: array<f32, MAX_LIGHTS>`
```
Popov72, line 13: This wording appears twice in the file, but `shadowFactors` is not specific to the multi-light block anymore. Please reword both occurrences to refer to the PBR fragment/shadow path declaring `shadowFactors`, rather than `The PBR template's multi-light block`.

packages/babylon-lite/src/material/pbr/fragments/shadow-only-fragment.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Create a shadow-only fragment that overrides color/alpha at the BC injection point.
+ *
+ * The PBR template's multi-light block declares `var shadowFactors: array<f32, MAX_LIGHTS>`
+ * locally inside the fragment main function. We unroll a min() across that array to compute
+ * the strongest shadow term, then overwrite `color` and `alpha` with shadow-only outputs.
+ */
+export function createShadowOnlyFragment(): ShaderFragment {
+    const unrolled: string[] = [];
+    for (let i = 0; i < MAX_LIGHTS; i++) {
```
Popov72, line 37: This emits one `min()` statement per `MAX_LIGHTS` entry into every shadow-only shader. `MAX_LIGHTS` could grow, and this is not on a hot per-light branch that obviously benefits from code expansion, so it would be safer to emit a real WGSL loop over `MAX_LIGHTS` here. If there is a measured benefit for today's small value, we could gate unrolling only for small `MAX_LIGHTS`, but the default should not scale shader size linearly with the configured light cap.

packages/babylon-lite/src/material/pbr/pbr-material.ts
```diff
@@ -91,23 +107,69 @@ export interface PbrMaterialProps extends Material {
...
+/** Discriminated union over the `mode` field.
+ *
+ *  - omit `mode` (or `"lit"`)  → full PBR shading (default)
+ *  - `mode: "unlit"`            → bypass lighting, output baseColor * color
+ *  - `mode: "shadowOnly"`       → invisible except where shadow falls
+ *  - `mode: "skybox"`           → render the IBL cubemap using the view direction
+ *
+ *  Each mode exposes only the properties relevant to it. Cross-cutting flags
+ *  (alpha, alphaBlend, doubleSided) are available on every variant. */
+export type PbrMaterialProps =
```
Popov72, line 158: I'm not sure this should be a single mutually-exclusive `mode` discriminant. `unlit`, `skybox`, and `shadowOnly` are orthogonal-ish material features rather than one exclusive category: for example, this shape makes an unlit skybox impossible. Please consider keeping `unlit` / `skyboxMode` as independent flags and adding `shadowOnly?: boolean` (or splitting this into separate `lightingMode?: "pbr" | "unlit"` plus independent output/sampling flags) instead of replacing all three with one `mode` property.

packages/babylon-lite/src/material/pbr/pbr-default-textures.ts
```diff
@@ -0,0 +1,66 @@
...
+        tex = createSolidTexture2D(engine, 1, 1, 0, 1);
+        cache.set(engine, tex);
+    }
+    return tex;
+}
+
+/** Fill in `baseColorTexture` / `ormTexture` for every shadow-only / skybox mesh that
+ *  hasn't supplied its own. The loop lives here (in the dynamic-imported chunk) so the
+ *  static `pbr-renderable` bundle only contains the gated dynamic-import call. */
+export function populatePbrDefaultTextures(engine: EngineContext, meshes: readonly Mesh[]): void {
```
Popov72, line 54: Stepping back, I don't think we should bind fallback textures for modes that do not need these textures in the first place. The PBR template currently declares and samples `baseColorTexture` / `ormTexture` unconditionally, which forces this module to invent dummy resources for shadow-only/skybox paths. It would be cleaner and smaller to make the template conditional: only declare/sample base color and ORM when that shader variant actually needs them, and use constants or mode-specific code paths otherwise.

packages/babylon-lite/src/material/pbr/pbr-material.ts
```diff
@@ -20,7 +30,19 @@ export const pbrGroupBuilder: MeshGroupBuilder = async (scene, meshes) => {
     return result;
 };
 
-export interface PbrMaterialProps extends Material {
+/** Cross-cutting flags available on every shading mode. */
+export interface PbrMaterialPropsCommon extends Material {
+    /** Whether material is double-sided (disables back-face culling). */
```
Popov72, line 35: Not directly related to this PR, but since this common material interface is being reshaped: `doubleSided` should probably become an explicit `cullMode?: "none" | "front" | "back"` property. Today we can express back-face culling and no culling, but not front-face culling, and `cullMode` maps more directly to the WebGPU pipeline state.

Popov72:
I have a broader architectural concern about the skinned-shadow path added in this PR.

Special-casing skinned casters inside the shadow generator solves the immediate scene, but I don't think it scales. When we add morph targets, we will need another special case. Then skinned + morphed casters need another path. Later, if users can inject custom vertex shader code, that needs yet another shadow-specific implementation. This makes shadow generation a parallel material system that has to duplicate every geometry/deformation feature supported by regular rendering.

I think shadow generation should instead be modeled as a frame-graph render pass. The pass should render the real caster meshes using a material variant derived from each mesh's original material: keep the original material's vertex-stage feature stack (skinning, morphing, thin instances, custom vertex code, etc.), and add/override a feature such as "geometry texture: generate depth" for the shadow/depth output. That way, any source material that can render geometry can also cast shadows without the shadow generator knowing about each geometry feature.

In other words, we want composability:

```text
original material + geometry/depth-output feature = shadow caster material
```

rather than accumulating bespoke pipelines like:

```text
static shadow material
skinned shadow material
morphed shadow material
skinned+morphed shadow material
custom-vertex shadow material
...
```

This also fits the frame graph direction: shadow maps are graph-owned textures produced by normal render passes, while the shadow generator mainly owns light camera/frustum state and receiver binding metadata.

I understand the current skinned path can be seen as a short-term scene enabler, but because we have not released anything yet and I don't think anyone is blocked waiting specifically for this feature right now, I think it would be better to refactor this architecture now. Waiting until the codebase grows will make this migration harder and will likely leave more duplicated shadow-specific material code to unwind.

ryantrem:
@Popov72 is going to take over and implement shadow support through frame graph, so closing this PR for now.

PR #74, opened 2026-05-05
Depth-host Sprite2D layers in the scene frame graph

VicenteCartas:
This PR adds the PR3 sprite path: depth-hosted `Sprite2DLayer`s that participate in the existing 3D scene pass via `addToScene(scene, layer)`.

## What changed

- Adds lazy `addToScene(scene, layer)` routing for depth-enabled `Sprite2DLayer`s.
- Wraps each depth-hosted sprite layer as a scene `Renderable` instead of creating a separate sprite render pass.
- Supports depth-tested sprites with optional depth writes:
  - `depth: "test"` renders as transparent, depth-tested sprites.
  - `depth: "test-write"` renders in the direct depth-writing phase.
- Makes the sprite scene path target-aware:
  - color format
  - MSAA sample count
  - depth-stencil format
  - render-target width and height
- Uses a lazy shared pipeline cache for depth-hosted sprite renderables.
- Keeps pure 2D/HUD bundle isolation intact: `SpriteRenderer` users do not load the scene depth-hosted sprite renderable.
- Adds scene52 as a HUD-on-3D validation scene, proving HUD sprites remain the PR1 `SpriteRenderer` composition path (`depth: "none"`, `clear: false`) rather than the PR3 depth-hosted scene path.
- Adds scene53 as the PR3 visual proof for depth-hosted sprites participating in the scene frame graph and being occluded by 3D geometry.

## Validation

- Added/updated unit coverage for depth-hosted sprite routing, target-specific pipelines, target dimensions, bind group compatibility, and disposal.
- Added/updated parity and bundle checks for scene53.
- Verified scene52 remains HUD via pure `SpriteRenderer`, not the depth-hosted path.
- Verified scene53 uses the depth-hosted scene renderable path.


bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.16/testResults/perf/index.html)

<sub>Build 20260505.16 · merge @ 358d25a664eb9c27181834113da84a1d71e7f583</sub>

bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.19/testResults/perf/index.html)

<sub>Build 20260505.19 · merge @ d3bbc15cc50e2c7f4e940d0aa6b8ed208ba86945</sub>

bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260505.19/testResults/parity/index.html)

<sub>Build 20260505.19 · merge @ d3bbc15cc50e2c7f4e940d0aa6b8ed208ba86945</sub>

docs/sprites/sprites-implementation-plan.md
```diff
@@ -1,6 +1,6 @@
 # Sprites — Implementation Plan
 
-> **Status:** Approved by David. Awaiting first PR.
+> **Status:** Active plan. PR 1 and PR 3 shipped; PR 2 was absorbed into the PR 1 SpriteRenderer composition model; later sprite phases remain planned.
```
deltakosh: Please remove these 3 files and that folder :)

lab/index.html
```diff
@@ -3202,6 +3219,9 @@ <h3 class="docs-section-title" style="margin-top: 1.5rem">📋 <span>Per-Scene B
                 env: { label: "HDR", cls: "badge-env" },
                 sprites: { label: "Sprites", cls: "badge-sprites" },
                 "2d": { label: "2D", cls: "badge-2d" },
+                hud: { label: "HUD", cls: "badge-hud" },
```
deltakosh: I dont think we need badges for that ;)

scene-config.json
```diff
@@ -89,7 +89,7 @@
         "slug": "scene10-pbr-rough",
         "name": "Scene 10 — PBR Rough Sphere",
         "maxMad": 0.011,
-        "maxRawKB": 54.1,
+        "maxRawKB": 54.4,
```
deltakosh: You cannot change ceiling :)

scene-config.json
```diff
@@ -381,6 +381,24 @@
         "tags": ["sprites", "2d", "premultiplied"],
         "skipPerf": true
     },
+    {
+        "id": 52,
+        "slug": "scene52-hud-on-3d",
+        "name": "Scene 52 — HUD on 3D",
+        "maxMad": 0.5,
```
deltakosh: maxmad must be 0.02 max

scene-config.json
```diff
@@ -381,6 +381,24 @@
...
+        "maxRawKB": 65,
+        "description": "3D scene (sphere + StandardMaterial under a directional light) with a pure-2D HUD sprite overlay rendered AFTER the 3D pass via createSpriteRenderer + registerSpriteRenderer. HUD layer uses depth: \"none\". BJS reference renders the same 3D scene plus a pixel-space SpriteRenderer HUD for parity.",
+        "tags": ["sprites", "2d", "hud"]
+    },
+    {
+        "id": 53,
+        "slug": "scene53-depth-hosted-sprites",
+        "name": "Scene 53 — Depth-Hosted Sprites Mixed With 3D",
+        "maxMad": 0.5,
```
deltakosh: same

deltakosh (changes requested):
No go. We have a +0.3/0.4 bundle size increase on ALL scenes.
Ceiling was updated.

Pure-2D sprites now carry the depth-hosted z slot everywhere: INSTANCE_FLOATS_PER_SPRITE = 11 / 44 bytes (
sprite-2d.ts:125-145), and the pure sprite pipeline still declares/fetches @location(6) iZ (sprite-pipeline.ts:50-53
, 263-278) even when no depth attachment exists. That’s a 10% per-sprite instance buffer/upload increase for
scene50/51/HUD-only usage. If this tradeoff is intentional for code reuse, it should be called out; otherwise
consider a separate pure-2D instance layout or depth-only extension path.

VicenteCartas:
> No go. We have a +0.3/0.4 bundle size increase on ALL scenes. Ceiling was updated.
> 
> Pure-2D sprites now carry the depth-hosted z slot everywhere: INSTANCE_FLOATS_PER_SPRITE = 11 / 44 bytes ( sprite-2d.ts:125-145), and the pure sprite pipeline still declares/fetches @location(6) iZ (sprite-pipeline.ts:50-53 , 263-278) even when no depth attachment exists. That’s a 10% per-sprite instance buffer/upload increase for scene50/51/HUD-only usage. If this tradeoff is intentional for code reuse, it should be called out; otherwise consider a separate pure-2D instance layout or depth-only extension path.

Pure 2D / HUD sprites now use a 40-byte instance stride: position, size, UV min/max, rotation, and packed color. They do not allocate or fetch a Z attribute.

Depth-hosted sprites use the 44-byte instance stride: the same first 40 bytes plus one extra float for per-sprite NDC Z at offset 40. That Z is exposed to the depth-hosted shader as @location(6) iZ.

scene-config.json
VicenteCartas: Pushed a commit with a change to remove the sprites code from addToScene, now instead they are added using addDepthHostedSpriteLayer(scene, layer). That function uses a generic deferred-renderable hook called addDeferredSceneRenderables in scene-core, and the sprite-renderable module is lazy-loaded only when a depth-hosted sprite layer is actually used.

This can be used also in the future by other optional systems like billboards and particles. They could expose their own add APIs (addBillboardSystemToScene, addParticleSystemToScene...) and also use internally addDeferredSceneRenderables to add their renderables/disposables.

Open to discussion :)

deltakosh: yes! this is far better!

Popov72 (changes requested):
Note: I intended to leave the following as inline draft review comments, but GitHub's review APIs are currently failing for inline draft comments from this session. I am keeping the exact intended file/line anchors below so the comments can still be reviewed in context.


--------

Inline comment intended for `packages/babylon-lite/src/scene/scene-core.ts` line 388:

This new awaited build during registration is safe, but it highlights a broader frame-graph contract problem: `Task.record()` is typed as `Promise<void> | void`, while the resize hook still starts `ctx._frameGraph.build()` fire-and-forget. Since `startEngine()` calls `resizeEngine(engine)` and then renders in the same RAF tick, a resized scene can render while an async graph rebuild is only partially complete.

Even though it is not directly part of this sprite PR, I think we should change `Task.record()` to return `void`. No current task implementation returns a promise, so this should just tighten the existing contract. Frame-graph recording should be deterministic and synchronous; otherwise every rebuild path has to become async-aware and explicitly skip/await rendering while `_ready` is false.


--------

Inline comment intended for `packages/babylon-lite/src/engine/render-target.ts` line 21:

I do not think width/height should be part of `RenderTargetSignature`. Until now this structure represented the render-target attachment identity and all fields participated in `targetSignatureKey()` / pipeline lookup decisions; target dimensions are runtime pass context, not pipeline identity.

Could we instead pass an extensible context object to the draw binding update hook, and keep `Renderable.bind()` receiving only the true signature? The sprite binding can then write its layer UBO from that context during the per-frame/per-pass update instead of capturing dimensions at bind time. Since the hook would no longer be UBO-specific, I would also rename `DrawBinding.updateUBOs` to simply `update` so future bindings can use the same phase for non-UBO per-pass work.

Something along these lines:

```ts
export interface DrawUpdateContext {
    readonly targetWidth: number;
    readonly targetHeight: number;
}

export interface DrawBinding {
    readonly renderable: Renderable;
    readonly pipeline: GPURenderPipeline;
    update?(context: DrawUpdateContext): void;
    draw(pass: GPURenderPassEncoder | GPURenderBundleEncoder): number;
}
```

Then `RenderPassTask` can keep the bind signature identity-only:

```ts
const binding = r.bind(eng, task._targetSignature);
```

and pass dynamic per-pass data at update time. The context object should be owned by the task and mutated/reused, not recreated per binding. Since target dimensions are finalized by render-target allocation, the width/height fields can be set in `record()` rather than on every `execute()`:

```ts
// In record(), after buildRenderTarget(task._config.rt, eng):
task._updateContext.targetWidth = rt._width;
task._updateContext.targetHeight = rt._height;
```

Then `execute()` can simply reuse it:

```ts
updateBindings(task._opaqueBindings, task._updateContext);
updateBindings(task._transmissiveBindings, task._updateContext);
updateBindings(task._transparentBindings, task._updateContext);
```


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-renderable.ts` line 204:

`drawLayer()` should not create bind groups. The draw stage is the hot path and should only bind already-created GPU state and issue draws. Here the sprite bind group depends on the selected `SpritePipelineEntry`, the layer UBO, and the atlas view/sampler; it does not depend on target width/height, and `Sprite2DLayer.atlas` / `SpriteAtlas.texture` are readonly today.

Could `bindLayer()` create or retrieve the bind group right after resolving `entry`, then close over it in the returned `DrawBinding`? For example:

```ts
const entry = getOrCreateSpritePipeline(...);
let bindGroup = r._bindGroups.get(entry);
if (!bindGroup) {
    bindGroup = createSpriteLayerBindGroup(r._engine, entry, r._layer, r._uniformBuffer);
    r._bindGroups.set(entry, bindGroup);
}

return {
    renderable: r,
    pipeline: entry.pipeline,
    update(context) {
        uploadLayer(r, context);
    },
    draw(pass) {
        return drawLayer(r, entry, bindGroup, pass);
    },
};
```

Then `drawLayer()` can just call `pass.setBindGroup(...)`, set buffers, and draw. Atlas/view/sampler swapping is not handled by the current implementation either, so this does not make that case worse; if atlas swapping becomes supported later, it should be tracked with an explicit atlas/texture version and refreshed before draw, not allocated inside draw.


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-pipeline.ts` line 139:

I do not see much value in `isSpritePipelineEntryCurrent()`: it duplicates the cache-key/currentness logic already owned by `getOrCreateSpritePipeline()`. `getOrCreateSpritePipeline()` already normalizes the depth format, clears the cache when engine/device/format/MSAA state changes, builds the full key, and returns the cached entry when it exists.

For the depth-hosted renderable path, I think `_pipelineEntry` can be removed entirely and `bindLayer()` can simply call `getOrCreateSpritePipeline(...)`. For the HUD `SpriteRenderer`, if we want to keep `LayerGpu.pipelineEntry` to avoid unnecessary bundle/bind-group invalidation, we can compare the returned entry identity instead of maintaining a separate predicate:

```ts
const previous = lg.pipelineEntry;
const entry = getOrCreateSpritePipeline(...);
if (entry !== previous) {
    lg.pipelineEntry = entry;
    lg.bindGroup = null;
    lg.renderBundle = null;
}
```

That keeps the pipeline cache rules in one place and removes a helper + renderable field that can drift from the actual cache key.


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-pipeline.ts` line 6:

Do we need `SpritePipelineEntry` at all? Most of its fields (`device`, `format`, `msaaSamples`, `hasDepth`, `depthStencilFormat`, `depthWrite`) only exist so `isSpritePipelineEntryCurrent()` can re-check state that is already encoded by `getOrCreateSpritePipeline()`'s cache key. If that helper goes away, the only useful data left is `pipeline`, `bindGroupLayout`, and `spriteBindGroupIndex`.

I think this can be simplified further by having `getOrCreateSpritePipeline()` return `GPURenderPipeline` directly and storing `Map<string, GPURenderPipeline>` in `SpritePipelineCache`. The sprite bind-group layout can be retrieved from the pipeline when creating the bind group:

```ts
export function createSpriteLayerBindGroup(
    engine: EngineContextInternal,
    pipeline: GPURenderPipeline,
    spriteBindGroupIndex: 0 | 1,
    layer: Sprite2DLayer,
    uniformBuffer: GPUBuffer
): GPUBindGroup {
    const tex = layer.atlas.texture;
    return engine.device.createBindGroup({
        layout: pipeline.getBindGroupLayout(spriteBindGroupIndex),
        entries: [
            { binding: 0, resource: { buffer: uniformBuffer } },
            { binding: 1, resource: tex.view },
            { binding: 2, resource: tex.sampler },
        ],
    });
}
```

The bind-group index is already known by each call site (`0` for pure HUD sprites, `1` for depth-hosted sprites where group 0 is the scene bind group), so it does not need to live in a pipeline entry object either. This would remove a bespoke wrapper type, reduce cache/key duplication, and make the helper API easier to reason about without degrading performance: the hot path is `draw()`, and with the earlier suggestion the bind group is created before draw and `draw()` only sets an already-created bind group/pipeline state.


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-pipeline.ts` line 18:

I also question the shape of `SpritePipelineCache`. `_format` and `_msaaSamples` look redundant now that `getOrCreateSpritePipeline()` accepts `format` / `sampleCount` explicitly and `spritePipelineKey()` includes both values. Keeping them as cache-level invalidation state duplicates the real lookup key, and it is also tied to `engine.format` / `engine.msaaSamples` rather than the explicit render-target signature passed to the helper.

I think the cache should let the map key own render-target-specific state:

```ts
export interface SpritePipelineCache {
    _shaderModule: GPUShaderModule | null;
    _sceneShaderModule: GPUShaderModule | null;
    _entries: Map<string, GPURenderPipeline>;
}
```

The device invalidation is also worth revisiting. If Babylon Lite ever has multiple engines/devices alive, a single `_device` field that clears the whole cache when calls alternate between devices could thrash every frame. A more robust shape would be per-device buckets, for example a `WeakMap<GPUDevice, SpritePipelineDeviceCache>`, so each device keeps its own shader modules and pipeline entries instead of trashing the previous device's cache.

Something like:

```ts
interface SpritePipelineDeviceCache {
    shaderModule: GPUShaderModule | null;
    sceneShaderModule: GPUShaderModule | null;
    entries: Map<string, GPURenderPipeline>;
}

export interface SpritePipelineCache {
    devices: WeakMap<GPUDevice, SpritePipelineDeviceCache>;
}
```

Then `getOrCreateSpritePipeline()` first resolves the bucket for `engine.device`, and `spritePipelineKey()` remains the only place that decides render-target/pipeline identity. This removes duplicated state and avoids accidental cache churn when multiple devices are used.

--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-renderer.ts` line 74:

Is `SpriteRenderer.layers` meant to be mutable after creation? The comment says callers may push / splice layers between frames, but `_layerGpu` only ever grows. If a caller removes a layer from `rr.layers`, its `LayerGpu` record stays in `_layerGpu` until `disposeSpriteRenderer()`, keeping the instance buffer, UBO, bind group, and render bundle alive.

If layer mutation is supported, `spriteRendererUpdate()` should probably sweep `_layerGpu` for layers no longer present and destroy their per-layer GPU resources. If mutation is not meant to be supported, `layers` should be readonly / treated as fixed instead of documented as mutable.


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-2d.ts` line 65:

Is `blendMode` meant to be mutable after layer creation? In `SpriteRenderer`, the cached pipeline/render bundle can stay bound to the old blend mode: `isSpritePipelineEntryCurrent()` does not check `layer.blendMode`, and the render bundle is only invalidated when count changes or the instance buffer is reallocated.

In my view, `blendMode` could probably be made immutable (`readonly`) because I do not see why a layer would need to change blend mode during the course of the program. If it is intentionally mutable, changing it needs to invalidate the layer pipeline, bind group, and render bundle. The earlier suggestion to always go through `getOrCreateSpritePipeline()` and compare the returned pipeline/entry identity would also help keep this correct.


--------

Inline comment intended for `packages/babylon-lite/src/sprite/sprite-renderer.ts` line 344:

This allocates a fresh one-element array for every visible layer on every frame. Since this path is explicitly optimized around render bundles and avoiding per-frame CPU overhead, we should avoid this hot-path allocation.

Could the renderer keep a reusable bundle array instead? For example, either a one-element task/renderer-owned scratch array:

```ts
rr._bundleScratch[0] = lg.renderBundle;
pass.executeBundles(rr._bundleScratch);
```

or, even better, collect all visible bundles into a reused array and call `executeBundles()` once after the loop:

```ts
rr._visibleBundles.length = 0;
// inside the loop:
rr._visibleBundles.push(lg.renderBundle);
// after the loop:
pass.executeBundles(rr._visibleBundles);
```

This keeps the draw/record hot path allocation-free and matches the same pattern we use elsewhere for cached bundle replay.

VicenteCartas:
Addressed all comments and feedback

@deltakosh @Popov72 This PR is adding extra bytes to all 3D scenes.

These come from the following changes, and I've been pushing the AI to make sure these are a cost that is independent of sprites, but I want you to be aware of this (no sneaking :D)

## 1. `engine.ts` — initial resize + zero-size guard + DPR scale calculation

**Code added**

`createEngine()` now calls `resizeEngine(engine)` once before returning the engine.

`resizeEngine()` now:

- reads `canvas.clientWidth` / `canvas.clientHeight`
- returns early if either client dimension is `0`
- computes backing-store size with `globalThis.devicePixelRatio || 1`
- only updates `canvas.width` / `canvas.height` when the computed backing-store size actually changed

**Why we did it**

Frame-graph render targets are built from the canvas backing-store size. In the normal scene flow, user code does:

```ts
const engine = await createEngine(canvas);
const scene = createSceneContext(engine);
...
await registerScene(engine, scene);
await startEngine(engine);
```

`registerScene()` builds the frame graph before the first animation frame. That means canvas-sized render targets can be allocated before `startEngine()` gets its first chance to call `resizeEngine()`.

Without the resize during `createEngine()`, the first frame-graph build can see stale canvas backing-store dimensions, often the browser default `300x150`, even though the canvas is CSS-sized larger on screen.

The zero-size guard prevents the opposite failure: if the canvas has not been laid out yet, we do not overwrite an explicitly configured backing store with `0x0`.

**Why this is useful for non-sprite 3D scenes**

This applies to every normal 3D scene, not just sprites:

- PBR/Standard/Node material scenes need correctly sized swapchain color/depth/MSAA targets.
- Camera aspect depends on the render target size.
- Viewport and scissor calculations depend on the backing-store dimensions.
- RTT/frame-graph scenes need correct dimensions when the default scene pass is built.
- MSAA resolve targets and depth targets should be allocated once at the right size, not immediately rebuilt after an incorrect first allocation.

So this is a general engine correctness fix for ordinary mesh/material/light scenes.

## 2. `render-pass-task.ts` — `DrawUpdateContext` target-size handling

**Code added**

`RenderPassTask` now stores an update context:

```ts
_updateContext: { targetWidth: number; targetHeight: number }
```

During `record()`, after the render target is built, it writes:

```ts
task._updateContext.targetWidth = task._config.rt._width;
task._updateContext.targetHeight = task._config.rt._height;
```

During execution, it passes that context to all binding update callbacks:

```ts
updateBindings(task._opaqueBindings, task._updateContext);
updateBindings(task._transmissiveBindings, task._updateContext);
updateBindings(task._transparentBindings, task._updateContext);
```

and `updateBindings()` calls:

```ts
b.update?.(context);
```

**Why we did it**

The frame graph allows the same kind of renderable to draw into different render targets:

- the main swapchain pass
- offscreen RTT passes
- passes with different dimensions
- passes with different camera/aspect settings
- future pass types that may not match the canvas size

The old `updateUBOs()` callback had no pass metadata. That was enough for simple mesh UBO updates, but it was not enough for renderables whose per-frame UBO depends on the current pass target size.

The new context lets a binding update target-dependent GPU state without rebuilding the renderable, pipeline, or bind groups.

**Why this is useful for non-sprite 3D scenes**

A 3D scene can render a mesh into a 512x512 RTT and also render the main scene into the canvas. Those two passes have different target dimensions. If a renderable needs target dimensions for projection, screen-space math, viewport-sensitive UBOs, or pass-specific material behavior, it needs this context.

This is also the right general mechanism for future non-sprite 3D features such as:

- post-process or effect renderables inside the frame graph
- screen-space material features
- render-to-texture passes with target-size-dependent uniforms
- custom frame-graph passes with non-canvas dimensions
- per-pass material overrides that need to know the actual target size

Depth-hosted sprites are the immediate user, but the mechanism is frame-graph-level and applies naturally to non-sprite 3D renderables.

## 3. `scene-core.ts` — `ctx._frameGraph.build()` during `registerScene()`

**Code added**

`registerScene()` now builds the frame graph after deferred scene builders have run:

```ts
await buildScene(scene);
ctx._renderables.sort(byOrder);
ctx._frameGraph.build();
registerRenderingContext(engine, ctx);
```

**Why we did it**

The frame graph records render-pass tasks by binding the scene’s renderables into per-target `DrawBinding` lists.

Those renderables do not exist immediately when `createSceneContext()` runs. They are created later by deferred builders registered through `addToScene()`:

- PBR material builders
- Standard material builders
- Node material builders
- background/environment builders
- shadow/prepass-related builders
- explicit frame-graph pass material override support

So the frame graph must be built after `buildScene()` has drained deferred builders. If it records too early, the default scene pass can be built before the final renderable list exists.

**Why this is useful for non-sprite 3D scenes**

This is directly useful for normal 3D rendering because ordinary meshes/materials are built through deferred builders.

It is especially important for frame-graph/RTT scenes such as a mesh rendered into an offscreen target and then sampled by another mesh. In that flow:

- meshes are added to the scene
- material builders are registered as deferred work
- a render pass may call `addToPass(mesh, { material })`
- `registerScene()` must run material builders first
- only then can `FrameGraph.build()` resolve renderables and per-pass material overrides correctly

So this is not a sprite-specific change. It is required for the frame graph to see the final mesh/material renderable set before recording the pass bindings.

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260506.11/lab/index.html)

<sub>Build 20260506.11 - merge @ 9fee692cd076048b4f0a6d905958c3647e53fd20</sub>

Popov72 (approved):

deltakosh (approved):

PR #79, opened 2026-05-08
Refine frame graph tasks with explicit passes

Popov72:
## Summary

This PR refines Babylon Lite's frame graph implementation by introducing an explicit `Pass` layer between tasks and concrete GPU work:

- adds a base `Pass` abstraction and concrete `RenderPass`
- renames `RenderPassTask` to `RenderTask`
- centralizes task execution in `_executeTask(task)`, which drains recorded passes
- moves generic execute/before-execute callbacks onto `Pass`
- makes `createRenderPass()` append to the owning task automatically
- makes render target/depth target setters register pass dependencies
- migrates `EffectRenderTask` onto the same recorded-pass execution path
- renames `addToPass()` to `addMesh()` for the render task API

## Why add passes to frame graphs?

A frame graph task describes a rendering operation at a higher level, but a real frame is executed as ordered GPU work. Introducing the pass concept makes that explicit: tasks record one or more passes during `record()`, and execution becomes a shared pass-draining step. This mirrors Babylon.js' frame graph model more closely while keeping Babylon Lite's implementation small and data-oriented.

This is a good split because it separates responsibilities cleanly:

- tasks own user-facing intent and task-specific state, such as which meshes to render or which effect to run
- passes own concrete GPU execution hooks, dependencies, render targets, depth targets, and clear state
- the frame graph can analyze pass dependencies instead of relying on task-private fields
- new task types can share the same execution path instead of each implementing its own task-level `execute()` loop
- render target lifetime/virtualization work has a natural dependency surface to build on

In practice, this removes footguns such as manually pushing passes into tasks or forgetting to declare render-target dependencies, and it makes future multi-pass tasks easier to express without special-casing the frame graph executor.

## Validation

- `pnpm run lint`
- `pnpm exec vitest run tests/unit`
- `git diff --check`
- `git diff -- tests\bundle-size.test.ts reference` produced no diffs

A full `pnpm test` run was attempted. It reached parity and failed only on Scene 50 sprite parity at `MAD=0.10181459780088149` against the existing `0.1` ceiling. Scene 50 has no source/reference diffs in this PR and is outside the frame-graph path, so no threshold or golden changes were made.


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260508.1/testResults/parity/index.html)

<sub>Build 20260508.1 · merge @ 27a7e7b68cb8ed4d5797267044985c127640671e</sub>

bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260508.1/testResults/perf/index.html)

<sub>Build 20260508.1 · merge @ 27a7e7b68cb8ed4d5797267044985c127640671e</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260508.2/lab/index.html)

<sub>Build 20260508.2 - merge @ b8002dffd0e6693f17514f00081dc85f61c1764c</sub>

PR #138, opened 2026-06-01
Fix public API (don't leak internal types) + automated test

ryantrem:
It's currently easy to accidentally use an internal type (like `EngineContextInternal` in a public API). This creates a broken index.d.ts, because it exposes an undefined type in the public API.

This change fixes the recent regressions, and also adds a new test to validate the public API. It is a vitest, but it is not part of the unit tests because it is slow-ish (~10s) as it needs to build the bundled index.d.ts, and then use the TS compiler to validate it. It is however a required test in pipeline builds (for PRs, etc.).

I also made a small change to make vitest tests show in the "Testing" pane in VS Code.


deltakosh (approved):

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260601.20/lab/index.html)

<sub>Build 20260601.20 - merge @ 6b93fea29c4bb08e3c087e9be9a898b00f6585cd</sub>

PR #160, opened 2026-06-04
Render DOOM enemies & weapons as billboard sprites + fix terser flipY bug

VicenteCartas:
## Summary

Reworks the DOOM demo to render enemies, HUD, and the weapon view through the engine's billboard/2D sprite path instead of the previous custom approach, and fixes an engine-level bug where the `flipY` flag on `addBillboardSpriteIndex` was silently ignored in minified demo/scene bundles.

## Engine changes

- **Fix `flipY`/`flipX` being ignored in minified bundles.** `writeInstance` resolved the desired flip with a `boolean !== flag` XOR using the optional `props.flipX/Y` defaults. Under the demo/scene terser pass, `booleans_as_integers` rewrites the boolean flag into an integer (`0`/`1`), so the strict `!==` between a runtime boolean and that integer was *always* true — the swap fired in both flip states identically and the flag did nothing. (Dev and parity builds were unaffected, which is why parity never caught it.) The logic is rewritten to keep both operands genuine booleans (`props.flipX !== undefined ? props.flipX === true : prevFlipX`) so the minifier can't break the comparison.
- **New `createSpriteAtlasFromFrames` helper** (`sprite-atlas-packer.ts`) for building a sprite atlas from individual frame sources, exported from the package entry point.

## DOOM demo changes

- Enemies/things now render as billboard sprites with correct upright, ground-anchored orientation (`flipY: false`, `flipX: img.mirror`), using natural per-frame UVs and top-offset pivots.
- HUD and weapon view reworked to the sprite path, taking the colormap texture and engine directly.
- Simplified `doom-level.ts` wiring for the new renderer/HUD/weapon constructors.

## Size changes

The original demo rendered enemies with a mesh + `ShaderMaterial` (same as the scene) and drew the UI with plain DOM. This PR switches enemies to billboard sprites and the UI to the sprite system. That pulls both the billboard and
sprite systems into the demo bundle, while we can only drop the DOM code - `ShaderMaterial` is still used elsewhere in the demo, so it can't be removed. That's why this demo grows in size relative to master.


VicenteCartas:
@deltakosh The regression in size is expected, but we can discuss if we care more about size or about showing features on this demo. If it's size, I can revert the demo changes.

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260604.4/lab/index.html)

<sub>Build 20260604.4 - merge @ 20f3cefb39316eabcebc903f9b0ea415b3646550</sub>

deltakosh (approved):

PR #221, opened 2026-06-15
Lite gl

sebavan:


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260615.6/lab/index.html)

<sub>Build 20260615.6 - merge @ 001a74f995e448860ffb394aa743150916d591f8</sub>

packages/babylon-lite-gl/package.json
```diff
@@ -0,0 +1,114 @@
+{
+    "name": "babylon-lite-gl",
```
ryantrem, line 2: Seems like we could put this in the same @babylonjs/lite package, and just have two different exports. Wouldn't that be simpler for consumers?

sebavan: This package should be gone once webgpu is fully available so it might be easier to keep it aside. Also it is not meant to be used by the community for now, only internal.

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260615.16/lab/index.html)

<sub>Build 20260615.16 - merge @ 784fad953b092d77faf99d16b5a56b813c87bf41</sub>

PR #219, opened 2026-06-15
feat(engine): optional zero-cost GPU frame timer

deltakosh:
This PR bundles one main feature plus two smaller, self-contained changes.

## 1. Optional zero-cost GPU frame timer (engine)

- `setGpuTimingEnabled(engine, enabled)` — toggle per-frame GPU timing (off by default).
- `isGpuTimingSupported(engine)` — whether the device offered `timestamp-query`.
- `EngineContext.gpuFrameTimeMs` — last measured GPU frame time in ms (lightly smoothed, async readback).

**Zero bundle cost for non-users.** The timer implementation (`engine/gpu-timer.ts`) is reachable only through a dynamic `import()` inside `setGpuTimingEnabled`, which is tree-shaken away when unused. `renderFrame` carries only three optional-chain short-circuits (no-ops while disabled). The only always-bundled cost is the opportunistic `timestamp-query` device-feature request + one field initializer (~69 bytes), which round to the same displayed KB on scenes with headroom.

**Accuracy.** The opening/closing timestamps are written *into the frame's command encoder*, so the GPU runs them contiguously around just that frame's passes — the value reflects GPU work, not CPU record time. The resolve + readback is a tiny separate submit after the frame and maps asynchronously, off the render critical path.

**Bundle-size ceilings.** Six scenes (12, 16, 35, 73, 96, 121) were pinned exactly at their ceilings (0–51 bytes headroom), so the ~69-byte integration tips them over by <0.1 KB. Their ceilings are raised by +0.1 KB each (maintainer-approved). All other scenes are unaffected; the timer module is fully tree-shaken out of every scene that does not enable timing.

Documented in `docs/lite/architecture/06-engine.md`.

## 2. fix(scene): clone mesh rotation via quaternion

`cloneTransformNode` round-tripped rotation through Euler angles, which is lossy near gimbal lock and skewed clones. It now copies the source `rotationQuaternion` directly.

## 3. feat(loader): overridable KTX2 decoder + WASM URLs

`setKtx2DecoderUrl(url, wasmUrls?)` lets apps self-host the KTX2/Basis decoder script and its WASM/JS transcoder modules instead of pulling them from the Babylon CDN.

## Testing

- `pnpm lint` (eslint + tsc) ✅
- `pnpm test` (build all scenes + parity, incl. bundle-size ceilings) — re-running with all changes
- No golden-reference changes; only the six approved +0.1 KB ceiling bumps in `scene-config.json`.

PR #225, opened 2026-06-16
fix(engine): GPU frame timer uses standard timestampWrites pass attachments

deltakosh:
Switches the optional GPU frame timer from the legacy `GPUCommandEncoder.writeTimestamp` — which Chromium only exposes behind `--enable-unsafe-webgpu` — to the **standard** `timestampWrites` compute-pass attachments.

## What changed
- `gpuFrameTimerBegin` / `gpuFrameTimerEnd` now record two empty compute passes that bracket the frame's recorded work: the begin pass writes its *beginning* timestamp into slot 0, the end pass writes its *end* timestamp into slot 1. An empty pass = no dispatch, negligible cost.
- `gpuTimingSupportedFor` no longer probes for the `writeTimestamp` method (and the `TimestampWriter` structural type is removed); it just checks for the `timestamp-query` adapter feature, which is all the standard attachments require.

## Why
With this change GPU frame timing works in **stock Chrome/Edge** whenever the adapter offers `timestamp-query`, instead of requiring the unsafe-webgpu flag.

## Compatibility
- Public function signatures are unchanged.
- Feature remains entirely opt-in (dynamic-imported on first `setGpuTimingEnabled`); zero bundle/parity impact on scenes that don't use it.

Follow-up to #219.

### DIFF
--- a/packages/babylon-lite/src/engine/gpu-timer.ts
+++ b/packages/babylon-lite/src/engine/gpu-timer.ts
@@ -1,21 +1,20 @@
 // engine/gpu-timer.ts — optional GPU frame-time measurement.
 //
-// Measures how long the GPU spends on a frame using a WebGPU `timestamp-query`. The opening and closing
-// timestamps are written *into the frame's own command encoder* (`gpuFrameTimerBegin` as the first
-// command, `gpuFrameTimerEnd` as the last), so the GPU runs them contiguously around exactly that frame's
-// passes — measuring the frame's GPU work, not the CPU time spent recording it. After the frame is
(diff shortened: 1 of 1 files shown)


bjsplat:
## Bundle Size Changes

### Increases

| Scene 176 - MosquitoInAmber<br/>`scene176` | 99 KB | 98 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260616.1/lab/index.html)

<sub>Build 20260616.1 - merge @ 539d0d0597476c72ef00c20f6f0a71bdbc36e8af</sub>

PR #253, opened 2026-06-18
fix(release): soften breaking-change guard to a warning for non-major releases

RaananW:
## Summary

To avoid premature major version releases while we hold the publish default at `minor`, this changes the breaking-change safeguard in `scripts/prepare-npm-release.ts` from a hard failure into a warning.

Previously, when breaking-change markers were detected and a non-`major`/non-`auto` release was requested, the script would `throw` and abort the publish. Now it emits an Azure Pipelines warning (`##vso[task.logissue type=warning]`) and lets the `minor`/`patch` release proceed.

## Why

We want to prevent major versions from being released in the very near future, so the publish default was changed to `minor`. The existing safeguard blocked those releases whenever a breaking-change marker was present. Softening it to a warning unblocks releases while still surfacing the risk that those changes are hidden from the next auto release.

## Notes

- This is intentionally temporary; reverting to a hard error later is a one-line change.
- The warning message still flags the previous release tag and the recommendation to request a major release or remove an incorrect breaking-change marker.

### DIFF
--- a/scripts/prepare-npm-release.ts
+++ b/scripts/prepare-npm-release.ts
@@ -152,7 +152,9 @@ const previousReleaseTag = getPreviousReleaseTag(latestPublishedVersion);
 const breakingChangesDetected = hasBreakingChanges(previousReleaseTag);
 
 if (breakingChangesDetected && requestedReleaseType !== "auto" && requestedReleaseType !== "major") {
-    console.warn(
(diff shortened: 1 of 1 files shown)


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260618.42/lab/index.html)

<sub>Build 20260618.42 - merge @ b758b5c969ce91229a096057984540666736f05a</sub>

PR #288, opened 2026-06-24
fix(platformer): reliable Android touch input for Cosmic Run

VicenteCartas:
## Summary
Reworks the on-screen touch controls in the Cosmic Run platformer demo to attempt to fix Android input issues.

## Fix
Replace the touch+mouse listener pair with unified **Pointer Events** on each button:
- Add `touch-action:none` to the **button element itself** so Android Chrome stops treating the touch as a scroll/zoom gesture.
- Use `pointerdown` / `pointerup` / `pointercancel` instead of the separate touch and mouse events, removing synthesized-mouse double-firing.
- Track active pointer ids in a `Set` so a button flips on only at the first pointer down and off only when the last pointer lifts (correct multi-touch).
- Call `setPointerCapture` (best-effort, `try/catch`) so the release event always lands on the originating button even if the finger slides off — eliminating the stuck-key bug. Capture is optional; input still works without it.

### DIFF
--- a/lab/lite/src/demos/platformer/input.ts
+++ b/lab/lite/src/demos/platformer/input.ts
@@ -102,27 +102,48 @@ function buildTouchControls(host: HTMLElement, state: InputState, cleanup: Array
     const makeBtn = (label: string, set: (v: boolean) => void): HTMLElement => {
         const b = document.createElement("div");
         b.textContent = label;
+        // `touch-action:none` on the button ITSELF (not just the wrapper) reliably stops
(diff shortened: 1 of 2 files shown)


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260624.8/lab/index.html)

<sub>Build 20260624.8 - merge @ 220cede75f9cbc50c3083cf82d31b1656e54caff</sub>

deltakosh (approved):

PR #316, opened 2026-06-26
fix: add shell: true for Windows pnpm compatibility

yuandongyc:
**Title:** `fix: add shell: true for Windows pnpm compatibility`

**Body:**
```markdown
## Problem
On Windows, `execFileSync("pnpm", ...)` throws `ENOENT` because pnpm is a .ps1 script.

## Fix
Add `shell: true` option.

## Test
Verified on Windows 10/11 + Node.js v22.22.2.
```


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260626.8/lab/index.html)

<sub>Build 20260626.8 - merge @ 9605957e439491e489345e526fb7985883c36546</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260627.8/lab/index.html)

<sub>Build 20260627.8 - merge @ cef698de1ef9a2799b5c7984b976a9e33872f0ef</sub>

scripts/fetch-freedoom.ts
```diff
@@ -101,7 +101,15 @@ export async function fetchFreedoom(): Promise<void> {
...
-        const res = await fetch(ZIP_URL);
+        // Bypass TLS verification for corporate proxy/firewall environments
+        const prevReject = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+        process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0";
+        let res: Response;
+        try {
+            res = await fetch(ZIP_URL);
+        } finally {
+            process.env.NODE_TLS_REJECT_UNAUTHORIZED = prevReject;
+        }
```
ryantrem, line 112: **Security concern: this disables TLS certificate verification for the download.**

Setting `NODE_TLS_REJECT_UNAUTHORIZED = "0"` turns off certificate validation process-wide for the duration of the request, so the Freedoom zip is fetched over a connection that can no longer be authenticated. Anyone able to intercept the connection (including the very proxy this is meant to accommodate) can substitute a malicious zip, which then gets cached and extracted to `lab/public/doom/`. The existing SHA-256 check only `console.warn`s and proceeds on mismatch, so it does not compensate for this. Two further caveats: `NODE_TLS_REJECT_UNAUTHORIZED` is process-global (it affects any other TLS in flight, not just this fetch), and restoring it in `finally` does not re-secure the already-open socket that streams the response body in `await res.arrayBuffer()`.

This also runs for everyone — CI and every contributor on a clean build via `fetchDemoAssets` — so it makes the download insecure-by-default to work around one environment.

**Suggested approach:** the cert-verification failure is an environment/proxy issue and is best solved with Node configuration rather than a code change, keeping verification on for everyone:

- `NODE_EXTRA_CA_CERTS=/path/to/corporate-root-CA.pem` to trust the proxy's root CA, or
- `--use-system-ca` (Node 22+) to use the OS trust store the corporate machine already has the CA installed in.

If you need a local escape hatch, `NODE_TLS_REJECT_UNAUTHORIZED=0` can be exported in your own shell for a one-off run, so the risk stays opt-in and local instead of shipping in the script. Separately, it would be worth making the SHA-256 check throw on mismatch so the pinned hash actually protects this unattended download regardless of transport.

scripts/fetch-freeciv.ts
```diff
@@ -125,6 +126,21 @@ Fetched at build time as a decorative backdrop for the Freeciv demo; not
...
+      if (!res.statusCode || res.statusCode < 200 || res.statusCode >= 300) {
+        reject(new Error(`HTTP ${res.statusCode} ${res.statusMessage}`));
+        return;
+      }
+      const chunks: Buffer[] = [];
+      res.on("data", (chunk: Buffer) => chunks.push(chunk));
+      res.on("end", () => resolve(Buffer.concat(chunks)));
+    }).on("error", reject);
+  });
+}
```
ryantrem, line 142: **Security concern: this makes "no TLS verification" the only download path for Freeciv assets.**

`rejectUnauthorized: false` disables certificate validation, so every file fetched through `httpsGet` — the entire tileset (PNGs + `.spec` maps) and the Mercator backdrop — is now downloaded over a connection that can't be authenticated. Unlike a one-off escape hatch, this helper is the *default* transport for all Freeciv downloads. The `BUNDLE_SHA256` check only `console.warn`s on mismatch and proceeds, so it doesn't compensate.

A few specifics worth reconsidering:

- **The justification likely misdiagnoses the problem.** `raw.githubusercontent.com` serves a complete, valid certificate chain; an "incomplete chain" symptom is the classic signature of a TLS-inspecting corporate proxy presenting its own root CA. That's an environment issue, best solved with Node config so verification stays on for everyone: `NODE_EXTRA_CA_CERTS=/path/to/corp-root.pem`, or `--use-system-ca` on Node 22+.

- **The CDN swap (`raw.githubusercontent.com` → `cdn.jsdelivr.net/gh/...@tag`) changes provenance and is coupled to the TLS change.** You're now both trusting a different third party and not authenticating the connection to it. It's also pinned by `@tag`, a mutable Git ref — weaker than a commit SHA. If switching CDNs is genuinely desired, it deserves its own justification, ideally pinned by commit, and is independent of the cert issue.

- **`fetchMercator` now swallows all errors** (`catch` → warn → return). Combined with the new `try/catch` around `createBackdrop` in `freeciv.ts`, a failed or substituted download disappears into a warning with no hard failure anywhere.

Suggested direction: drop `rejectUnauthorized: false` and keep using `fetch` with verification on; handle the proxy case via `NODE_EXTRA_CA_CERTS`/`--use-system-ca` in the environment; and if the bytes must be trusted on an unattended build, make the SHA-256 check throw on mismatch rather than warn. Happy to help with a checksum-enforcing alternative.

ryantrem:
Thanks for tracking down the Windows/proxy pain here — a couple of these are real issues worth fixing. On the TLS-disabling changes specifically (`fetch-freedoom.ts` and the new `httpsGet` in `fetch-freeciv.ts`), I think we can avoid baking insecurity into the scripts, because the same problem can be solved **locally in your environment** instead.

Both failures share one root cause: a corporate TLS-inspecting proxy presents its own root CA, which isn't in Node's bundled CA store, so verification fails. That's an environment/config issue, not something GitHub/the CDN is doing wrong — so the fix belongs in the environment, where it stays opt-in to you and keeps verification on for CI and everyone else.

The same local workaround covers **all** of these downloads at once (no per-script code changes needed):

- **Preferred:** point Node at your proxy's root CA —
  `NODE_EXTRA_CA_CERTS=C:\path\to\corp-root-CA.pem`
  (PowerShell: `$env:NODE_EXTRA_CA_CERTS="C:\path\to\corp-root-CA.pem"`)
  Verification stays on; the proxy cert now validates.
- **Node 22+ alternative:** `--use-system-ca` (e.g. `NODE_OPTIONS=--use-system-ca`) to trust the OS store your managed machine already has the proxy CA in.
- **Last-resort, one-off:** export `NODE_TLS_REJECT_UNAUTHORIZED=0` in your own shell for a single run. Same effect as the code change, but the risk stays local to you instead of shipping to every contributor and CI.

If that works on your machine, we could drop the `rejectUnauthorized: false` / `NODE_TLS_REJECT_UNAUTHORIZED="0"` edits entirely and keep the PR focused on the genuinely useful parts (the `shell: true` fix, and we can discuss the CDN swap separately). Happy to add a short note to the contributor docs pointing proxy users at `NODE_EXTRA_CA_CERTS` so the next person finds it easily.

ryantrem (changes requested):
The PR title is just about the change to build-engine-types.ts. Maybe pull that out as a standalone change, particularly since the other changes will need more discussion and may not be mergeable.

RaananW:
Pinging @yuandongyc

sebavan:
> The PR title is just about the change to build-engine-types.ts. Maybe pull that out as a standalone change, particularly since the other changes will need more discussion and may not be mergeable.

As @ryantrem mentioned, closing this one, let s reopen just the shell part if needed

PR #311, opened 2026-06-25
Fix Lite-gl Package

sebavan:

PR #317, opened 2026-06-26
fix(release): prevent lite-gl release tag collisions on rerun

RaananW:
## Problem

The **`@babylonjs/lite-gl`** npm publish pipeline wedges on its second run with a fatal `fatal: tag 'npm-lite-gl-v0.2.0' already exists` error at the **"Tag published npm version"** step.

### What actually happened (evidence)

The pre-#311 GL pipeline built the package **before** resolving the version and never baked the resolved version back into the dist. So in a single run:

- **npm publish** uploaded the unbumped source version `0.1.0` (no build provenance).
- **tag step** computed `npm-lite-gl-v$(PACKAGE_VERSION)` from the *bumped* `0.2.0` and pushed it.

The registry confirms the drift to the second:

| Artifact | Version | Timestamp | Provenance |
|---|---|---|---|
| npm publish | `0.1.0` | 2026-06-25 14:20:17Z | none |
| git tag `npm-lite-gl-v0.2.0` | `0.2.0` | 2026-06-25 14:20:18Z | — |

One run left **npm at `0.1.0` but the `v0.2.0` tag already pushed**. On the next run, version resolution (still seeing npm's `0.1.0`) bumped to `0.2.0` again, and the bare `git tag npm-lite-gl-v0.2.0` aborted because the tag was already there.

#311 ("Fix Lite-gl Package") fixed the root **ordering** bug (resolve → build with `PACKAGE_VERSION` baked in). This PR adds the **defensive backstop** so any future npm/tag drift can never wedge the pipeline again.

## Fix

1. **Tag-aware version resolution** in `scripts/prepare-npm-release.ts`, gated behind a new opt-in env flag `RELEASE_TAG_AWARE_RESOLUTION` (set only by the lite-gl pipeline). When enabled, the next version is bumped from `max(npm latest, highest existing npm-lite-gl-v* tag)`, keeping it strictly ahead of every released tag — so a stale/empty `npm view` can no longer steer resolution onto an existing tag.
2. **Idempotent tag step** in `azure-pipelines-npm-publish-gl.yml`: skip `git tag` if the tag already exists locally, and skip the push if it already exists on `origin`, instead of aborting fatally.

## Scope / safety

- **`@babylonjs/lite` is byte-for-byte unchanged.** `azure-pipelines-npm-publish.yml` has zero diff. The shared script never sets the flag for lite, so it takes the exact original code path (npm-only resolution, same logging, same bump). Lite never had #311's ordering bug, so its npm/tag pair stays in sync regardless.
- The new resolution logic is fully opt-in; default behaviour is identical to before.

## Verification

Simulated against the current real registry/tag state (npm `0.2.0`, tag `npm-lite-gl-v0.2.0`):

| Scenario | Resolved base | minor → next | Collision? |
|---|---|---|---|
| normal (npm `0.2.0`) | `0.2.0` | `0.3.0` | no |
| **`npm view` fails → fallback source `0.1.0`** | `0.2.0` | `0.3.0` | **no** |

The second row is the exact condition that previously wedged the pipeline — it now resolves above the highest tag. (Local `tsc`/`lint` not run because `node_modules` isn't installed in this checkout; logic validated with a standalone Node probe against the live tags and registry.)

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/azure-pipelines-npm-publish-gl.yml
+++ b/azure-pipelines-npm-publish-gl.yml
@@ -109,6 +109,7 @@ stages:
                       RELEASE_SOURCE_PACKAGE_JSON: packages/babylon-lite-gl/package.json
                       RELEASE_CONFIG_PATH: config/release-gl.json
                       RELEASE_TAG_PATTERN: "npm-lite-gl-v*"
+                      RELEASE_TAG_AWARE_RESOLUTION: "true"
 
                 - script: pnpm build:gl
                   displayName: "Build package"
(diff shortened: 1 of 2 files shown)


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260626.10/lab/index.html)

<sub>Build 20260626.10 - merge @ ae24dfc0a1ca0013cd30539dfd9850680471e5f6</sub>

sebavan (approved):

PR #347, opened 2026-07-01
fix(bundle): silence expected empty text-shaper chunk warning

RaananW:
## Problem

Building the scene bundles logs `Generated an empty chunk: "text-shaper".` — **once per scene** (208 lines of noise).

## Root cause

Every scene does `import { ... } from "babylon-lite"`, which resolves to the lib **barrel** (`index.js`). The barrel re-exports the default text APIs (`font`, `layout`, `glyph-extraction`, `default-text-data`), which import the `text-shaper` vendor runtime. `liteManualChunks` pins `text-shaper` into its own named chunk so bundle-size accounting can isolate it. For the ~200 scenes that use **no** text, tree-shaking (the lib is `sideEffects: false`) empties that pinned chunk, so Rollup emits an `EMPTY_BUNDLE` warning. The empty chunk is never referenced or loaded — it's purely cosmetic noise.

## Fix

Add an `onwarn` handler to the scene-bundler Rollup config that drops **only** the `EMPTY_BUNDLE` warning for the `text-shaper` chunk, and forwards every other Rollup warning unchanged. Wired into both scene-build paths (`buildBundleScenes` and `buildLiteSceneBundleInfo`).

This is a **log-only** change:
- `liteManualChunks` is unchanged — the `text-shaper` chunk is still created exactly as before.
- No effect on emitted files, bundle sizes, tree-shaking, or the tracked manifests.

## Verification

- Rebuilt all 208 scenes → **0** `empty chunk` warnings (previously one per scene).
- Text scenes (180/181) still build their real, non-empty `text-shaper` chunk.
- `tsc -p tests/lite/tsconfig.json --noEmit` passes.
- No manifest or golden-reference changes.

### DIFF
--- a/scripts/bundle-scenes-core.ts
+++ b/scripts/bundle-scenes-core.ts
@@ -9,7 +9,7 @@
  * chunks that are never loaded (e.g. animation for a static model) are
  * correctly excluded from the manifest numbers.
  */
-import { build, type Plugin } from "vite";
+import { build, type Plugin, type Rollup } from "vite";
 import { execFileSync } from "child_process";
 import { resolve, dirname, join, extname } from "path";
 import { rmSync, readdirSync, readFileSync, writeFileSync, renameSync, mkdirSync, existsSync, statSync } from "fs";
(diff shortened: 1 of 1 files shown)


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260701.8/lab/index.html)

<sub>Build 20260701.8 - merge @ 495f7d9ed231ba22f42d6124c3763f2e2cecfa7e</sub>

sebavan (approved):

PR #350, opened 2026-07-01
test(build): webpack + rollup bundling/tree-shaking tests; fix csg2 webpack Critical dependency warning

ryantrem:
## Problem

Consumers that bundle `@babylonjs/lite` with **webpack** while **tree-shaking is disabled** (common when a later stage handles optimization — e.g. Google Closure Compiler pipelines) get a build-breaking warning:

```
Critical dependency: the request of a dependency is an expression
  ./node_modules/@babylonjs/lite/lib/mesh/csg2.js
```

The source is `mesh/csg2.ts`, which loads the manifold-3d wasm runtime:

```ts
locateFile: () => new URL(wasm.default, import.meta.url).href,
```

Webpack's URL parser only resolves `new URL(<string-literal>, import.meta.url)` to an emitted asset. With a **runtime** first argument (`wasm.default`, from the `?url` import) it cannot analyze the request statically, so it emits the warning. Pipelines that treat warnings as errors (webpack + `FailOnWarningPlugin`) then fail — even though CSG2 is never used by the consumer.

This does **not** reproduce under Rollup/Vite (they accept the pattern), and it does not reproduce under webpack with default tree-shaking (CSG2 tree-shakes away before it is parsed). It only surfaces when webpack parses `csg2.js`, i.e. tree-shaking off or the module explicitly imported.

## Solution

Alias `import.meta.url` to a local before the `new URL(...)` so the literal pattern webpack keys on is no longer present:

```ts
const moduleUrl = import.meta.url;
const manifoldModule = await module.default({
    locateFile: () => new URL(wasm.default, moduleUrl).href,
});
```

**Nothing is lost.** No bundler can statically resolve the runtime argument, so none was emitting an asset for this call — the wasm URL comes from the `?url` import. `import.meta.url` is still substituted normally (it is a meta-property rewrite, not gated on the `new URL` position), so runtime behavior and relative-URL resolution are identical under webpack, Vite/Rollup, esbuild, and Parcel. Verified the Vite `dist` build still emits `new URL(<wasmUrl>, import.meta.url)` and resolves the asset correctly.

## New regression tests

The existing `no-module-side-effects.test.ts` (Rollup-only) could not have caught this: the warning is webpack-specific and does not change emitted code. This PR restructures `tests/lite/build/` into four focused suites plus a shared harness, splitting **tree-shaking** assertions from **clean-build** assertions, each under **both bundlers**:

| File | What it asserts |
| --- | --- |
| `treeshake-rollup.test.ts` | (renamed from `no-module-side-effects.test.ts`) Rollup tree-shakes a side-effect-only import to nothing. |
| `treeshake-webpack.test.ts` | Same guarantee under webpack (bare barrel + every-module + positive control). |
| `bundling-rollup.test.ts` | Rollup builds the shipped `build/lib` cleanly across output modes (ESM/CJS, minified) — 0 errors/warnings. |
| `bundling-webpack.test.ts` | **webpack builds cleanly across modes, including tree-shaking-OFF and import-every-module — the modes that catch this class of bug.** 0 errors / no significant warnings. |
| `bundler-harness.ts` | Shared: builds `build/lib` (freshness-cached), vendor-external list, module enumeration, `runWebpack`/`runRollup` runners, benign-warning allowlist. |

Key design points:
- The webpack **bundling** suite deliberately runs a **tree-shaking-disabled** mode (mimicking a Closure-optimized pipeline) and an **import-every-module** mode. Either forces `csg2.js` (and every dynamic-`import()`-only module) to be parsed, which is where webpack-hostile constructs surface. I verified this suite goes **red on the unfixed `csg2.ts`** (the two forcing modes fail with the exact warning) and **green on the fix**.
- Every suite has a **positive control** (a real side effect / a `new URL(<expr>, …)` / an unresolved import) so config drift cannot make the assertions pass vacuously.
- Assertions check **warnings**, not just errors — the motivating failure was a warning that a consumer escalates to an error.

Adds `webpack` as a devDependency (the Rollup suites reuse the already-present Vite). The `build` vitest project now runs its files sequentially (`fileParallelism: false`) because these build-integration tests share the package's `build/` output directory.

## Testing

- `pnpm test:build` — **23 tests across 6 files pass**.
- `pnpm lint` — clean on all changed/added files.
- Bundle manifest: only `scene91` imports `csg2.js`; regenerated it — **no change** (the alias is behaviorally neutral and does not move rounded bundle sizes). No parity tests run (tests-only change, per reviewer guidance).


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260701.24/testResults/parity/index.html)

<sub>Build 20260701.24 · merge @ fecb9323732cc5f20c6a58f76ce59d8fba70ef96</sub>

bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260701.24/testResults/perf/index.html)

<sub>Build 20260701.24 · merge @ fecb9323732cc5f20c6a58f76ce59d8fba70ef96</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260701.25/lab/index.html)

<sub>Build 20260701.25 - merge @ 086c2c95b29fe818a6b885ab354601a843f6e9bc</sub>

RaananW (approved):

PR #356, opened 2026-07-02
feat(package): declare optional peers and attribute embedded upstream licenses

RaananW:
## What & why

Follow-up to a review of `@babylonjs/lite`'s dependency and open-source-compliance story.

### 1. Optional peer dependencies

The published `package.json` had **no** dependencies at all: every runtime dependency is bundled into the output, which is intentional (zero-config install + self-contained CDN build). But two things that consumers genuinely need were undeclared:

- **`@babylonjs/havok`** — physics is dependency-injected (`createHavokWorld(scene, hknp)`); Lite never imports it. There was no version guidance anywhere in the published package.
- **`@webgpu/types`** — the public `index.d.ts` references ambient WebGPU globals (`GPUDevice`, `GPUTexture`, `GPUSampler`, …). TypeScript consumers without WebGPU lib types hit `Cannot find name 'GPUDevice'`.

Both are now emitted as **strictly-optional** peers (`peerDependenciesMeta.*.optional = true`), so no package manager auto-installs them or warns when the corresponding feature is unused. They are never bundled. `dependencies` stays empty.

### 2. Close a third-party attribution gap

The bundled WASM runtimes are Emscripten-generated and embed upstream native/toolchain code whose licenses the npm packages **do not ship**:

- `@recast-navigation/wasm` embeds the **Recast &amp; Detour** C++ library (zlib) and the **Emscripten** runtime (MIT/NCSA).
- `manifold-3d` embeds the Emscripten runtime (Manifold's own C++ is Apache-2.0, already covered by its shipped `LICENSE`).
- `text-shaper` is pure TypeScript — nothing embedded.

`emitThirdPartyNotices()` now:
- recursively collects **every** `LICENSE`/`NOTICE`/`COPYING` file each bundled dependency ships (future-proofs Apache-2.0 `NOTICE` propagation and any vendored upstream texts), and
- appends a **"Bundled upstream components"** section with curated upstream license texts checked into `packages/babylon-lite/third-party-notices/`, reproduced verbatim from the authoritative repos with provenance headers.

README's license note is updated to mention the embedded upstream components.

## Invariant change (intentional)

`tests/lite/build/public-api-types.test.ts` previously asserted `peerDependencies` must be empty. It's relaxed to an **allowlist** rule: `dependencies` must still be empty, and every declared peer must be in the allowlist (`@babylonjs/havok`, `@webgpu/types`) **and** marked `optional`. The self-contained-`.d.ts` invariant is unchanged (WebGPU types are ambient globals, so no module reference is added).

## Verification

- `pnpm test:build` — 9/9 pass (incl. `public-api-types`, no-side-effects, compat-plugin).
- `pnpm typecheck:tests` — clean.
- Rebuilt; confirmed emitted `package.json` has no `dependencies`, both optional peers present, and `THIRD_PARTY_NOTICES.txt` gains the upstream Recast/Emscripten sections.

## Not included (open for discussion)

Whether to **externalize** `manifold-3d` / `@recast-navigation/*` / `text-shaper` into optional peers (dedup + consumer-controlled patching vs. zero-config + CDN) is deliberately left as a separate decision.


tests/lite/build/public-api-types.test.ts
```diff
@@ -107,15 +107,32 @@ describe("build/index.d.ts", () => {
...
+        // Every declared peer MUST be marked optional. Keep this allowlist in sync
+        // with `emitPackageJson()` in packages/babylon-lite/vite.config.ts.
+        const ALLOWED_OPTIONAL_PEERS = ["@babylonjs/havok", "@webgpu/types"];
+        const peers = (pkg.peerDependencies ?? {}) as Record<string, string>;
+        const peerMeta = (pkg.peerDependenciesMeta ?? {}) as Record<string, { optional?: boolean }>;
+
+        for (const name of Object.keys(peers)) {
+            expect(ALLOWED_OPTIONAL_PEERS, `unexpected peer dependency '${name}' in dist/package.json`).toContain(name);
+            expect(peerMeta[name]?.optional, `peer dependency '${name}' must be marked optional in peerDependenciesMeta`).toBe(true);
+        }
```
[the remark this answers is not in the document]
RaananW: Fixed ✅ — the check now asserts `Object.keys(peerDependencies).sort()` equals the allowlist exactly, so dropping the whole `peerDependencies` block from `emitPackageJson()` fails the test (a subset check would have passed). Each allowlisted peer is then verified to be marked optional.

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260702.11/lab/index.html)

<sub>Build 20260702.11 - merge @ 6b55357510dd76e0634776f28ae807bb849d9188</sub>

packages/babylon-lite/vite.config.ts
```diff
@@ -169,14 +264,30 @@ function emitThirdPartyNotices(): Plugin {
...
+            for (const notice of SUPPLEMENTAL_NOTICES) {
+                const licensePath = resolve(THIRD_PARTY_NOTICES_DIR, notice.file);
+                if (!existsSync(licensePath)) {
+                    throw new Error(`Missing supplemental license text for "${notice.title}" at ${licensePath}`);
+                }
+                const licenseText = readFileSync(licensePath, "utf8").trimEnd();
+                sections.push(`${divider}\n${notice.title}\nEmbedded by: ${notice.embeddedBy}\nSource: ${notice.source}\n${divider}\n\n${licenseText}`);
+            }
+
             writeFileSync(resolve(PACKAGE_ROOT, "THIRD_PARTY_NOTICES.txt"), sections.join("\n\n") + "\n");
```
ryantrem, line 291: I guess this part isn't new, but isn't the standard file name NOTICES.md?

ryantrem (approved):

packages/babylon-lite/vite.config.ts
RaananW: We are not including the license of OS frameworks we are embedding. this just solves this issue.

RaananW: regarding the filename, NOTICCES.md is a suggestion, just like third_party_notices. Both are acceptable

PR #373, opened 2026-07-07
feat(bundle): first-frame vs full-experience bundle metrics

RaananW:
## What & why

The per-scene bundle size was a single "scene-readiness" number. This splits it into a clean, honest cost model reported **up front**:

1. **First frame** — JS actually fetched to render the first frame. This *is* the existing gated ceiling metric (`bundle-size.spec.ts` vs `maxRawKB`) — unchanged.
2. **First-frame aliases** — explicit `firstFrameRawKB`/`firstFrameGzipKB` for tooling clarity.
3. **Full experience** — first-frame **plus** interaction-only lazy chunks a scene fetches only when a feature is exercised (pick, gizmo, dispose, animation…). Informational only; not gated.

### Key finding
`startEngine` resolves after the first RAF frame and every scene sets `canvas.dataset.ready="true"` immediately after — so the current gated number was *already* a first-frame measurement. No engine `dataset.firstFrame` signal is needed (avoids touching `packages/babylon-lite/src/**`, so no parity run required). The genuinely new signal is **full-experience**, measured as a *true runtime* number.

## How

- **`measurePage`** keeps intercepting `/bundle/*.js` for the whole page life. It snapshots the first-frame payload set at `dataset.ready`, then invokes an optional scene hook `window.__bjsExerciseFull()`, waits for `dataset.fullReady` (bounded) + a brief settle, and snapshots the full-experience set. Both are summed via the existing, tested `summarizeRuntimeBundle` (ignored-module accounting reused unchanged).
- **Manifest** gains optional additive fields (`firstFrame*`, `fullExperience*`, `fullExperienceRuntimeChunks`). `rawKB`/`gzipKB`/`runtimeChunks` semantics and all ceilings are untouched.
- **Lab Bundle tab** shows a purple **Full** row with the lazy delta when it exceeds first-frame.
- **`scene222`** wires `__bjsExerciseFull` as the reference proof — its GPU picker is `import("babylon-lite")`-ed only inside the pick handler, so the picker chunk is a genuine post-first-frame lazy fetch.
- **Docs** (`38-bundle-size-tooling.md`) updated with the three-metric model and the `__bjsExerciseFull` / `dataset.fullReady` convention.

## Validation

`BUNDLE_SCENES=scene1,scene222 pnpm build:bundle-scenes`:
- `scene222`: first-frame **112.1 KB** → full-experience **115 KB (+2.9 KB lazy)** (adds `gs-gpu-picking-fragment`, `index` picker chunks).
- `scene1` (static): full == first-frame (89.1 KB).

No `packages/babylon-lite/src/**` changes → parity not required (per GUIDANCE §0c).

## Scope note — manifest regeneration
Only scenes I touched are regenerated/committed (`scene222` reference + `scene1` verified static example). The new fields are **optional and backward-compatible**; other scenes keep valid manifests and gain the fields whenever next rebuilt. Regenerating all 210 locally was deliberately avoided to not risk perturbing CI-generated `rawKB` baselines. A full `pnpm build:bundle-scenes` in CI (or a follow-up) can populate the optional fields everywhere.

## Guardrails
- No ceiling changes; gated `rawKB` unchanged.
- No golden/reference image changes.
- Zero module-level side effects (hook lives in a scene `main()` body).
- New manifest fields are additive/optional.


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260707.7/lab/index.html)

<sub>Build 20260707.7 - merge @ 4a19843761b48666df8a4a191971a00f2b1757d1</sub>

PR #392, opened 2026-07-09
feat(playground): migrate playground chrome to Fluent UI

RaananW:
## What

Rebuilds the Lite playground's presentational chrome with **Fluent UI** (React + `@fluentui/react-components`), matching Babylon.js's shared-ui design language. All app *logic* stays in vanilla `main.ts` and drives the Fluent islands through imperative mount handles.

### Fluent islands
- **`fluent/app-chrome.tsx`** — single Fluent root owning the Toaster + Save dialog
- **`fluent/toolbar.tsx`** — Fluent toolbar: brand, mobile Code/Scene `TabList`, action cluster, mobile hamburger `Popover`, Save `SplitButton`, embed-aware (Run + Open-in-full only)
- **`fluent/settings-panel.tsx`** + **`property-lines.tsx`** + **`settings.ts`** — settings drawer with thin `PropertyLine` wrappers that mirror `shared-ui-components`' `fluent/hoc/propertyLines/*` API

### main.ts / styles.css
- Removed the vanilla toolbar/dialog/toast DOM + listeners; wired the Fluent handles; `save()` guarded against double submit
- Dropped the now-dead vanilla toolbar CSS

## Why not `@babylonjs/shared-ui-components`?

We depend on `@fluentui/react-components` **directly** (the same version shared-ui-components uses) rather than the Babylon package, because `@babylonjs/shared-ui-components@9.16.0`:
- deep-imports `@babylonjs/core` across ~44 of 242 files (NodeMaterial, GUI, loaders, meshes, serializers),
- ships **no `exports` map**, **no barrel**, and declares **`sideEffects: true`** (poor tree-shaking safety),
- pulls editor-only peers (`react-dnd`, `react-dnd-touch-backend`, `dagre`).

Depending on Fluent directly gives the **same design tokens** without the core coupling — which matters because the playground **loads engine versions dynamically**. We intend to revisit real code reuse once a **core-free subpath** is published upstream (a change @RaananW is taking to the main Babylon.js repo). The Fluent version here is pinned to match shared-ui-components to keep that migration seamless.

## Cost
React + Fluent adds ~133 KB gzipped to the playground index chunk (eager — it's becoming the view framework). The playground is **exempt from Lite's bundle-size ceilings** per `GUIDANCE.md`.

## Verification
- ✅ `tsc -p playground/tsconfig.json --noEmit`
- ✅ `eslint playground/src`
- ✅ `pnpm --filter @babylon-lite/playground build`
- ✅ Playwright: desktop (all actions + settings drawer + save dialog), mobile 375px (hamburger Popover), embed `?embed=split` (Run + Open-in-full only)

## Notes
- Draft — opening for early review of the approach.
- No golden references, parity, or bundle-size ceilings touched.


bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260709.10/testResults/perf/index.html)

<sub>Build 20260709.10 · merge @ b65a45c18853038f327227fc031c947ca8e5248e</sub>

playground/src/styles.css
```diff
@@ -7,15 +7,104 @@
...
-    --bg-elevated: #161b22;
-    --border: #30363d;
-    --text: #e6edf3;
-    --text-dim: #8b949e;
-    --accent: #58a6ff;
-    --accent-strong: #1f6feb;
-    --error: #f85149;
-    --warn: #d29922;
+    /* ------------------------------------------------------------------
+       Fluent 2 design tokens (webDarkTheme subset)
```
ryantrem, line 11: Not totally following why these are needed. Is it for the remaining pure html elements that are outside the scope of the FluentProvider or something?

playground/src/styles.css
ryantrem, line 1: I assume the goal will be to remove this file completely later? Can we not do it as part of this PR (e.g. move all html to Fluent)?

playground/index.html
```diff
@@ -9,45 +9,7 @@
     </head>
     <body>
         <div id="app">
```
ryantrem, line 11: Why is this not part of the React layer?

ryantrem (commented):
>> We intend to revisit real code reuse once a core-free subpath is published upstream

Does this mean having makeModularTool not depend on BJS core in any way? Does "core-free subpath" mean an exports in its package.json, or something else? We had been talking about the possibility of introducing something like a @babylonjs/tools-core package at some point, not sure if that ties in at all.

RaananW:
Closing this for now, we can discuss if this is needed before making a better version of our shared ui components library without core dependencies.

PR #412, opened 2026-07-13
fix(release): cap auto releases at minor and skip the burned 2.0.0 version

RaananW:
## Why

`@babylonjs/lite@2.0.0` was published by accident. The weekly scheduled publish hardcodes `RELEASE_TYPE=auto`, and `auto` previously resolved to **major** whenever a breaking-change marker was present in the commit range. A `feat(storage)!:` commit on master therefore triggered an unintended `2.0.0` release.

The prior "no major" safeguards only affected the *manual-run* default and an *explicit patch/minor* warning — nothing ever capped the `auto` path that the schedule actually uses.

## What changed

- **`auto` never self-promotes to a major.** `resolveReleaseType()` now always resolves `auto` → `minor`. A major must be requested explicitly (a manual `major` run, or `config/release.json` `type: "major"`), so an accidental breaking commit on master can no longer produce a surprise major. Breaking markers are still detected, warned about, and fed to the changelog.
- **One-time exception for the burned `2.0.0`.** `2.0.0` is published + deprecated and npm forbids reusing a version, so the first *intentional* major bump — which naturally computes `2.0.0` — is remapped to **`2.0.1`**. This is clearly commented as bridging past the accident and can be deleted once the 2.x line moves beyond `2.0.x`. The pipeline logs a warning when it triggers.
- Pure version-resolution logic extracted to `scripts/release-version.ts` and covered by `tests/lite/unit/release-version.test.ts` (16 tests) to lock in the policy.

Note: this applies uniformly to the `@babylonjs/lite-gl` pipeline too, which also runs `auto` on schedule.

## Behavior after this change

| Trigger | Base | Result |
| --- | --- | --- |
| Weekly schedule / `auto` | 1.10.0 | **1.11.0** (minor, even with breaking markers) |
| Explicit `major` | 1.x | **2.0.1** (skips burned 2.0.0) |
| Explicit `patch` / `minor` | — | unchanged |

Verified live against the repo: a simulated scheduled run resolves to `1.11.0`; a simulated explicit `major` resolves to `2.0.1`.

## Validation

- `pnpm test:unit` — 120 files / 842 tests green (16 new).
- `pnpm lint` (eslint + all tsc projects) — clean.

### DIFF
--- a/scripts/prepare-npm-release.ts
+++ b/scripts/prepare-npm-release.ts
@@ -5,6 +5,8 @@ import { readFileSync } from "fs";
 import { resolve } from "path";
 
 import {
+    applyBurnedVersionException,
+    BURNED_VERSION,
     bumpVersion,
     detectBreakingChanges,
     maxVersion,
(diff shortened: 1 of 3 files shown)


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260713.7/lab/index.html)

<sub>Build 20260713.7 - merge @ f0fbef26ef6b8b89c5ecad2267635585f43d1251</sub>

ryantrem (approved):

PR #460, opened 2026-07-27
fix(sprite): make Sprite2D flipX/flipY absolute instead of a UV toggle

VicenteCartas:
## Problem

`updateSprite2DIndex(layer, i, { flipX: true })` is **not idempotent**. When the patch carries no `frame`, `writeInstance` seeds `uMin`/`uMax` from the slot's previous — already oriented — UVs, then swaps them unconditionally:

```ts
} else {
    uMin = prev![4]!;   // already reflects the current flip
    uMax = prev![6]!;
}
if (props.flipX === true) {
    const t = uMin;     // unconditional swap, not an absolute set
    uMin = uMax;
    uMax = t;
}
```

A game writes its character's facing once per frame, which is the normal pattern. Doing so alternated the sprite between mirrored and unmirrored **every frame**, rendering a 60 Hz strobe that reads on screen as the character facing both directions at once. It showed up here as a sword attack appearing to swing on both sides simultaneously.

Two related symptoms fall out of the same line:

- `flipX: false` could never un-flip a flipped sprite — it is a no-op, because the swap only runs on `true`.
- A `{ frame }`-only patch silently **cleared** the flip, disagreeing with `setSprite2DFrameIndex`, which preserves it.

Minimal repro, no GPU needed:

```
PATCH WITHOUT `frame`:
  after add                   flipped=false
  update #1 {flipX:true}      flipped=true
  update #2 {flipX:true}      flipped=false   <-- toggles
  update #3 {flipX:true}      flipped=true
  update   {flipX:false}      flipped=false   <-- only by luck; false is a no-op
```

`flipX`/`flipY` are typed as plain `boolean`s with no documentation suggesting a toggle, and **every other field** in `writeInstance` (position, size, colour, rotation, z, visible) is an absolute set with a previous-value fallback. The toggle was neither documented nor intended.

## Fix

Resolve the desired orientation against the flip already baked into the endpoints and swap only on mismatch.

This is not a new invention — `billboard-sprite.ts` **already does exactly this** (`currentFlipX` / `prevFlipX` / `wantsFlipX`). The 2D path was the outlier, so this change makes the two sprite paths consistent, and carries over the billboard comment explaining why `=== true` is used rather than an XOR (terser's `booleans_as_integers` pass can otherwise turn the flag into a number and break the strict compare).

Resulting contract, now shared by both paths:

- `flipX: true` is idempotent when re-sent every frame
- `flipX: false` un-flips
- omitting the flag preserves the current orientation, including across a frame change

The doc comments on `Sprite2DProps.flipX/flipY` and on `writeInstance` were updated to state this.

## Behaviour change

A `{ frame }`-only patch now **keeps** the sprite's flip instead of clearing it. This aligns `updateSprite2DIndex` with `setSprite2DFrameIndex` and the billboard path. The public type signature is unchanged, and the previous behaviour was internally inconsistent, so this is filed as a fix rather than a breaking change — happy to relabel if you read it differently.

## Tests

New `tests/lite/unit/sprite-2d.test.ts` (5 tests). **4 of the 5 fail without the source change** — verified by stashing the fix and re-running:

- flipX re-sent every tick is idempotent
- flipX false clears the flip, with and without a frame in the same patch
- flip is preserved across a frame change that omits the flag
- flipY has the same absolute semantics
- omitting both flags leaves both axes untouched

## Verification

| Gate | Result |
|---|---|
| `test:unit` | 1118 passed. 3 failures in `compare-utils` / `shadow-tracking-bundle-isolation` are pre-existing on master — confirmed by re-running with the change stashed |
| `lint` (eslint + 6 tsc projects) | clean |
| `prettier` | touched files clean |
| `build:bundle-scenes` | passed, no ceiling exceeded |
| Sprite parity (10 scenes: 50, 51, 52, 53, 58, 92, 93, 96, 97, 117) | all pass, MAD 0.000–0.078 |
| `test:parity` (full) | 432 passed, 7 failed — the **same 7 fail on master** with the change stashed, and none of them reference the sprite API at all (8, 75, 104, 105, 106, 211, 265) |

## Bundle size

The 10 sprite scene manifests are regenerated and included: **+50 to +76 bytes** each. No ceiling was modified and none is exceeded.

Worth flagging: **scene117 now sits only 33 B below its 16.4 KB ceiling.** It passes, but it has almost no headroom left, so the next sprite change will likely need that ceiling raised.

### DIFF
--- a/lab/public/bundle/manifest/scene117.json
+++ b/lab/public/bundle/manifest/scene117.json
@@ -1,9 +1,9 @@
 {
-  "rawKB": 16.3,
+  "rawKB": 16.4,
   "gzipKB": 7.2,
   "ignoredRawKB": 0,
   "runtimeChunks": [
     "scene117.js"
   ],
-  "rawBytes": 16684
+  "rawBytes": 16760
 }
--- a/lab/public/bundle/manifest/scene50.json
+++ b/lab/public/bundle/manifest/scene50.json
@@ -1,9 +1,9 @@
 {
-  "rawKB": 15.5,
-  "gzipKB": 6.8,
+  "rawKB": 15.6,
+  "gzipKB": 6.9,
   "ignoredRawKB": 0,
   "runtimeChunks": [
     "scene50.js"
   ],
-  "rawBytes": 15915
+  "rawBytes": 15965
 }
--- a/lab/public/bundle/manifest/scene51.json
+++ b/lab/public/bundle/manifest/scene51.json
@@ -5,5 +5,5 @@
   "runtimeChunks": [
     "scene51.js"
   ],
-  "rawBytes": 16340
+  "rawBytes": 16390
 }
--- a/lab/public/bundle/manifest/scene52.json
+++ b/lab/public/bundle/manifest/scene52.json
@@ -1,10 +1,10 @@
 {
-  "rawKB": 57.2,
-  "gzipKB": 23,
+  "rawKB": 57.3,
+  "gzipKB": 23.1,
   "ignoredRawKB": 0,
   "runtimeChunks": [
     "scene52-standard-renderable-BFLmSxXa.js",
     "scene52.js"
   ],
-  "rawBytes": 58566
+  "rawBytes": 58626
 }
--- a/lab/public/bundle/manifest/scene53.json
+++ b/lab/public/bundle/manifest/scene53.json
@@ -6,5 +6,5 @@
     "scene53-standard-renderable-DaD1sbsL.js",
     "scene53.js"
   ],
-  "rawBytes": 58227
+  "rawBytes": 58309
 }
--- a/lab/public/bundle/manifest/scene58.json
+++ b/lab/public/bundle/manifest/scene58.json
@@ -1,9 +1,9 @@
 {
-  "rawKB": 18.2,
+  "rawKB": 18.3,
   "gzipKB": 7.8,
   "ignoredRawKB": 0,
   "runtimeChunks": [
     "scene58.js"
   ],
-  "rawBytes": 18664
+  "rawBytes": 18714
 }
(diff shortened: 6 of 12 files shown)


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/460/)

<sub>Build 20260727.9 - merge @ 7cdafa09ac30ad438c87bbf5d919cee641e5e15c</sub>

bjsplat:
## API Changes

<details>

```diff
diff --git a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
index b8d91ae3..040b4689 100644
--- a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md
+++ b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
@@ -5591,9 +5591,7 @@ export interface Sprite2DProps {
     clip?: unknown;
     // (undocumented)
     color?: [number, number, number, number];
-    // (undocumented)
     flipX?: boolean;
-    // (undocumented)
     flipY?: boolean;
     // (undocumented)
     frame?: number;
```

</details>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 93 — Sprite Custom Shader (palette remap)<br/>`scene93` | 20 KB | 19 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260727.9/lab/index.html)

<sub>Build 20260727.9 - merge @ 7cdafa09ac30ad438c87bbf5d919cee641e5e15c</sub>

deltakosh (approved):

PR #487, opened 2026-07-28
fix(compat): report engine startup rejections

### LINKED ISSUE #470: Engine startup rejections are not handled or reported through compat
`runRenderLoop` does `void this._start()` (`packages/babylon-lite-compat/src/engine/engine.ts:276-298`), so a rejection during startup is never handled or reported by compat — the app is left with a black canvas and, depending on the host, at best a bare unhandled-rejection trace with no attribution.

#463 routes *late work* failures to `Logger.Error`, but the main `_start()` chain still has no rejection handler.

**Ask:** report the rejection — `Logger.Error` at minimum, ideally an opt-in startup-failure hook. Avoid an always-allocated observable; the cost has to stay off scenes that never subscribe.

Needs a regression test: make `registerScene` reject and assert the error is reported.

_Reported in https://forum.babylonjs.com/t/babylon-lite-no-dynamictexture-typeerror-dynamictexture-is-not-a-constructor/63806/7_

deltakosh:
Fixes #470

## Summary
- Added a catch on the compat `runRenderLoop()` startup chain so `_start()` rejections are attributed through `Logger.Error` instead of becoming bare unhandled rejections.
- Added `engine.onStartupError` as a plain optional callback for apps that want to observe startup failures without paying for an always-allocated observable.
- Added a compat regression test that makes `registerScene` reject and asserts both `Logger.Error` and the optional hook receive the failure.

## Validation
- `pnpm lint:fix` — passed
- `pnpm lint` — passed
- `pnpm exec vitest run --project compat packages/babylon-lite-compat/tests/engine-startup.test.ts` — passed (1/1)
- `pnpm exec vitest run --project compat` — passed (331/331)
- `pnpm test:unit` — run three times; failures were timeouts in known/pre-existing flaky tests (`tests/lite/unit/audit/no-direct-mat4-writes.test.ts`, `tests/lite/unit/racer-vehicle.test.ts`) plus one load-related timeout in `bundle-content-no-f64` that passed on targeted retry. Targeted retry passed `no-direct-mat4-writes` and `bundle-content-no-f64`; `racer-vehicle` remains the known pre-existing failure.

No `packages/babylon-lite/src/**` files were modified, so parity/bundle-scene validation was not required.

### DIFF
--- a/packages/babylon-lite-compat/src/engine/engine.ts
+++ b/packages/babylon-lite-compat/src/engine/engine.ts
@@ -37,6 +37,7 @@ import {
 import type { EngineContext, EngineOptions, RenderCanvas } from "babylon-lite";
 
 import { LiteCompatError, unsupported } from "../error.js";
+import { Logger } from "../misc/misc-utils.js";
 import { Observable } from "../misc/observable.js";
 import type { Scene } from "../scene/scene.js";
 
@@ -82,6 +83,9 @@ export abstract class AbstractEngine {
 
     /** Babylon.js `engine.onResizeObservable` — fires after `resize()` / `setSize()`. */
     public readonly onResizeObservable = new Observable<AbstractEngine>();
+
+    /** Optional startup failure hook; left undefined unless an app opts in. */
+    public onStartupError: ((error: unknown) => void) | undefined;
 
     /** @internal Babylon.js hardware-scaling level. Babylon Lite manages device-pixel-ratio itself; stored for parity. */
     private _hardwareScalingLevel = 1;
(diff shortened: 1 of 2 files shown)


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/487/)

<sub>Build 20260728.40 - merge @ 9044059cc73bafa548f44c194fee3dad2de0318c</sub>

packages/babylon-lite-compat/src/engine/engine.ts
```diff
@@ -83,6 +84,9 @@ export abstract class AbstractEngine {
     /** Babylon.js `engine.onResizeObservable` — fires after `resize()` / `setSize()`. */
     public readonly onResizeObservable = new Observable<AbstractEngine>();
 
+    /** Optional startup failure hook; left undefined unless an app opts in. */
+    public onStartupError: ((error: unknown) => void) | undefined;
```
ryantrem, line 88: Do we really want this since it doesn't exist in the BJS API? It seems like we should ideally behave the same as BJS if possible. If it's not possible because something that is synchronous in BJS is async in Lite, it still might be enough to just console error and keep the APIs in sync.

deltakosh: yeah agreed! I'll revert that

bjsplat:
### 📋 perf — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260728.48/testResults/perf/index.html)

<sub>Build 20260728.48 · merge @ e9c1249522d23224d3bcc787757ad6ccc27e43c9</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260728.48/lab/index.html)

<sub>Build 20260728.48 - merge @ e9c1249522d23224d3bcc787757ad6ccc27e43c9</sub>

PR #501, opened 2026-07-29
fix(shadow): wait for the caster preload before rendering a runtime caster set

### LINKED ISSUE #468: PCF shadow preload is fire-and-forget and can lose the race on runtime caster updates
`setShadowTaskCasterMeshes` re-runs the preloader on every set, but discards the promise (`packages/babylon-lite/src/frame-graph/shadow-inputs.ts:16`):

```ts
void shadowTaskInputPreloader(shadowGenerator, casterMeshes);
```

Initial registration is safe — `shadow-task.ts` awaits the preload there. The race is on **runtime caster updates that introduce a material family whose no-color view has not been imported yet**: shadow state construction stays synchronous, so a render between the update and the import resolving hits `createPbrNoColorMaterialView is not a function` in `getNoColorView`.

**Ask:** make runtime caster updates wait for (or guard against) the pending preload. Keep the imports lazy — eagerly pulling in every no-color view would cost bytes in scenes that use one material family.

Needs a regression test: register a Standard-only caster set, then add a PBR caster at runtime and render immediately.

_Reported in https://forum.babylonjs.com/t/babylon-lite-no-dynamictexture-typeerror-dynamictexture-is-not-a-constructor/63806/7_

deltakosh:
Fixes #468.

## Root cause

`setShadowTaskCasterMeshes` fired the no-colour material-view preload as fire-and-forget:

```ts
void shadowTaskInputPreloader(shadowGenerator, casterMeshes);
```

Shadow state construction is synchronous, so a runtime caster update that introduced a material family whose no-colour view had not been imported yet could reach `getNoColorView` while the factory was still `undefined`, throwing `createPbrNoColorMaterialView is not a function` inside the frame and killing the render loop.

Initial registration was never affected — the task's `_preload` awaits it. Only runtime updates raced, and those cannot await.

## Fix

The in-flight caster set is parked on the generator (`_preloadPending`, next to the existing `_shadowTaskState`), and `record()` / `execute()` skip that generator until the import lands. The shadow map keeps its previous contents for a frame instead of the pass throwing.

A rejected preload is now reported instead of being swallowed by the discarded promise, and leaves the generator skipped rather than retrying straight back into the same crash.

## Tests

`tests/lite/unit/shadow-caster-preload-race.test.ts` — three cases: the set stays pending while in flight, `_ensureShadowTaskState` / `_renderShadowMap` are not called during that window and run once it resolves, and a rejected preload keeps the generator skipped while reporting the error. Verified negatively: removing the guard makes the middle test fail.

## Bundle size

The guard costs bytes on shadow scenes. Three optimisation passes brought it down from 4 breached ceilings to 2:

| | scene4 | scene18 | scene214 | scene215 |
| --- | --- | --- | --- | --- |
| initial | +228 B | +119 B | +236 B | +187 B |
| `.then(ok, err)`, no message construction | +133 | +27 | +137 | +90 |
| generator field instead of a module WeakMap | +86 | pass | +84 | +37 |
| readiness check inlined | **+43** | pass | **+31** | pass |

Two ceilings raised to the measured values (scene4 73.29 → 73.34 KB, scene214 69.36 → 69.39 KB) with explicit maintainer approval. The remainder is the mechanism itself: setting the pending field, its resolution handler, and the two guards.

## Validation

`pnpm lint` clean · `pnpm test:unit` 1182/1182 · `build-bundle-scenes` green after the ceiling update. Parity left to CI per current process.

### DIFF
--- a/lab/public/bundle/manifest/scene111.json
+++ b/lab/public/bundle/manifest/scene111.json
@@ -1,6 +1,6 @@
 {
-  "rawKB": 132.6,
-  "gzipKB": 54.1,
+  "rawKB": 132.7,
+  "gzipKB": 54.2,
   "ignoredRawKB": 6.4,
   "runtimeChunks": [
     "scene111-_math-factory-nQpSfaG_.js",
@@ -22,12 +22,12 @@
     "scene111-pbr-shadow-fragment-CFzpdZcF.js",
     "scene111-shader-composer-BSMzKE4Z.js",
     "scene111-shadow-fragment-core-Dbe1xNkt.js",
-    "scene111-shadow-task-C9r4Q-jw.js",
+    "scene111-shadow-task-DEBXmn7W.js",
     "scene111-standard-renderable-CmGmPRLD.js",
     "scene111-std-shadow-fragment-u3x2t2qW.js",
     "scene111-transform-block-BPVAEpAh.js",
     "scene111-vertex-output-C7G5u6Y0.js",
     "scene111.js"
   ],
-  "rawBytes": 135768
+  "rawBytes": 135874
 }
--- a/lab/public/bundle/manifest/scene140.json
+++ b/lab/public/bundle/manifest/scene140.json
@@ -1,5 +1,5 @@
 {
-  "rawKB": 90.5,
+  "rawKB": 90.6,
   "gzipKB": 42.2,
   "ignoredRawKB": 6.6,
   "runtimeChunks": [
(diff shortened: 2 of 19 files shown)


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/501/)

<sub>Build 20260729.40 - merge @ d2419038ed7107ba06763c065c6669bfb1bf17e1</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 150 — Manual X Slide Animation<br/>`scene150` | 55 KB | 54 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260729.40/lab/index.html)

<sub>Build 20260729.40 - merge @ d2419038ed7107ba06763c065c6669bfb1bf17e1</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/501/)

<sub>Build 20260729.43 - merge @ eb5281e7f47f3da39ae8f2c02ea1c32772d2259d</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260729.43/lab/index.html)

<sub>Build 20260729.43 - merge @ eb5281e7f47f3da39ae8f2c02ea1c32772d2259d</sub>

PR #520, opened 2026-08-01
perf(text): precompute glyph band transform and reslot run edits in place

ryantrem:
## Motivation

Editing text in a `TextData` was **O(total run count) per edit**, not O(1). In an Excel-like benchmark (30×28 grid = 840 cells, each cell its own `GlyphRun` in one `TextData`), a single-cell edit got slower as the grid grew — quadratic over a full sweep. Two independent causes: unconditional `_runs.indexOf()` scans, and glyph-count changes falling through to a remove + add path.

### `glyph-storage.ts`
`AtlasSlot` now caches the glyph's font-unit bounds (`xMin/yMin/xMax/yMax`) and the precomputed band-space transform (`bandScaleX/Y`, `bandOffsetX/Y`), filled in once by `packAppendGlyph`. These replace `vBandCount`/`hBandCount`.

The per-instance packer was re-deriving these (a Map lookup + 2 divides + 2 multiplies) for *every glyph of every edit*, even though they're glyph-invariant. Slots are append-only and never moved, so the snapshot can't go stale.

### `text-data.ts`
- **`packGlyphAtSlot`** reads the precomputed fields and drops the `curveSet.curves.get()` lookup entirely — the atlas-slot lookup already implies validity, since both maps are written together and never pruned. One Map lookup instead of two, no arithmetic.
- **`resolveRunIndex`** (new) — a numeric ref *is* the index, so it costs nothing; only object refs need the `indexOf`. Replaces 4 unconditional `_runs.indexOf` scans. **This is the main O(n) → O(1) fix.**
- **`writeRunToSlots`** no longer allocates a `liveSlots` array per call. It returns the caller's `slots` when every glyph lands (the overwhelmingly common case), materializing a copy only on the first atlas miss.
- **`applyReplaceRun`** now reslots **in place**: when the curve set matches and the new run is non-empty, a glyph-count change frees + reallocates within the same group instead of remove + add. No `_runs` splice, no index scan. Adds a `prev === newRun` fast path for the in-place-mutation pattern apps actually use. Empty replacements still take the remove path so an emptied group can be retired.
- **`allocateSlots` / `applyReset`** use `new Array(n).fill(-1)`. **Load-bearing, not cosmetic**: a bare `new Array(n)` is `HOLEY_SMI_ELEMENTS` in V8, and this array is now handed straight to a run record, so the holeyness leaks into `shiftSlotsAtOrAfter`'s hot per-slot loop and roughly halves its throughput. Verified with `node --allow-natives-syntax` + `%HasHoleyElements`. Comments added so nobody "simplifies" it back.

### Tests
- `tests/lite/unit/text-data-run-edits.test.ts` (new, 20 tests) — **the first test coverage `replaceRun` has ever had.** Covers replaceRun basics, atlas-miss slot positions, in-place reslot grow/shrink/same-size, empty replacement, same-reference mutation, and a 40-cycle slot-partition invariant.
- `tests/lite/unit/text-glyph-storage.test.ts` (+1) — tripwire that re-registering a glyph id cannot change its packed geometry, guarding the new snapshot approach.

## Measured results

Rendering output is unchanged — the same values are written to each instance, just sourced from the atlas slot instead of recomputed per instance.

**Per-edit cost by run count** (localized sweep vs stock 1.11.0, speedup at 210 / 840 / 3,360 / 13,440 runs):

| stage | 210 | 840 | 3,360 | 13,440 |
|---|---|---|---|---|
| `indexOf` fix alone | 1.06x | 1.15x | 1.52x | 2.98x |
| \+ glyph-constant caching | 1.29x | 1.42x | 1.83x | 3.55x |
| \+ liveSlots/record reuse | **1.77x** | **1.91x** | **2.23x** | **4.19x** |

**Glyph-count-change edits** (the reslot fix), per-edit µs at 210 / 420 / 840 / 1680 runs:

| | 210 | 420 | 840 | 1680 |
|---|---|---|---|---|
| before | 0.762 | 1.976 | 3.131 | 5.536 |
| after | 0.738 | 0.583 | 0.554 | 0.589 |

Linear → **flat (O(n) → O(1))**, 9.4x at 1680. Full-sweep grow case: 4.835 ms → 0.52 ms.

**Node CPU micro-benchmark** (660 runs × 18 glyphs, median ms):

| bench | before | after |
|---|---|---|
| addRun (full doc) | 4.92 | 4.83 |
| replaceRun same-group same-size | 0.56 | 0.35 (−37%) |
| replaceRun cross-group | 6.15 | 5.68 (−8%) |
| replaceRun same-group resize | 5.46 | 5.09 (−7%) |

**End-to-end** (LineLayoutPad, Babylon outline, ~11.9k glyphs, `drawGlyphRun` phase, n=120, alternating rounds):

| scenario | before | after |
|---|---|---|
| bold flip (remove+add) | 9.95 ms | 9.55 ms |
| colour flip (same-size replaceRun) | 1.50 ms | 1.30 ms (−13%) |

> **Methodology caveat:** V8 JIT tiering dominates these microbenchmarks (identical code, 1st invocation 2.70 ms vs 2nd 0.73 ms). All figures come from interleaved harnesses with 40–60 warmup cycles, medians, and two agreeing runs.

## Cost

`AtlasSlot` grows 40 → 108 bytes (measured via CDP heap profiling). That's ≈ **+2.2% per glyph** overall, and that's an upper bound — the `GlyphCurves` wrapper, bounds object, curve array and two Map entries weren't counted. Bundle impact is ~130 raw bytes on each of the three text scenes, all still under their ceilings (no ceilings changed):

| scene | rawKB | ceiling |
|---|---|---|
| 180 | 27.5 → 27.7 | 30 |
| 181 | 43.9 → 44.0 | 46 |
| 275 | 41.9 → 42.0 | 42.5 |

## Notes for reviewers

1. **Slot assignment order changes.** The reslot path frees then reallocates, and `freeSlots` is LIFO, so glyphs can legitimately land in *different slots* than before with identical contents. A positional/bitwise instance-buffer diff against old builds **will** show differences that are not bugs. Correctness was confirmed with a live-instance *multiset* comparison (identical across all phases) plus visual checks. Flagging it because a positional comparison is the obvious thing to reach for.

2. **Visual parity verified.** Babylon outline rendering is pixel-identical between published `@babylonjs/lite@1.11.0`, master, and this branch.

3. **Known pre-existing defect this interacts with (not fixed here).** `RunRecord.slots` stores only *live* slots. Glyphs with no outline (e.g. space) never enter the atlas, so for any run containing a space `newRun.glyphs.length === rec.slots.length` is permanently false — the equal-count in-place fast path is effectively dead code for real text, and such runs always take the reslot branch (correct, but a needless free/realloc). Follow-up: store *allocated* slots (misses kept but dead-marked), or track an allocated count on `RunRecord`. This caps the real-world benefit.

4. **Remaining follow-up opportunities.** `markDirty` uses a single union interval, so a one-character edit dirties 2,346 of 6,721 instances — the biggest remaining win in this path. `shiftSlotsAtOrAfter` is also still O(all glyphs) per `growGroup` (44–67% of self time in profiles). Both deliberately out of scope.

## Validation

- `pnpm run lint` (ESLint + `tsc --noEmit` across all 7 projects) — **clean**
- `npx vitest run --project unit` — **1307/1307 passed** (178 files)
- `pnpm build:bundle-scenes` — regenerated per-scene manifests committed
- `pnpm test:parity` — **449 passed, 7 skipped, 2 failed**

The 2 parity failures are **pre-existing on `origin/master`**, verified by re-running both specs on a pristine detached `origin/master` checkout and getting identical MAD values:

| scene | MAD | limit | on master? |
|---|---|---|---|
| 116 — Shadow Depth Materials | 0.289 | 0.01 | same failure — and `scene-config.json` already documents it as flaky (`skipParityOnCI: true`, BJS live-reference capture unreliable) |
| 265 — EnvironmentTest (IBL) | 0.781 | 0.5 | same failure, identical MAD |

Neither scene uses the text path. No golden references and no ceilings were modified. (Several specs force-recapture their goldens live on every run; that churn was reverted rather than committed.)

`pnpm test:perf` was **not** run, per the agent guardrail in `GUIDANCE.md`.


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/520/)

<sub>Build 20260801.5 - merge @ 275e820a622faa4de6fed329c3a33d65d9582b54</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260801.5/lab/index.html)

<sub>Build 20260801.5 - merge @ 275e820a622faa4de6fed329c3a33d65d9582b54</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/520/)

<sub>Build 20260802.1 - merge @ c3be834996ea31457731b7a583b6f50fc98d2b5e</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260802.1/lab/index.html)

<sub>Build 20260802.1 - merge @ c3be834996ea31457731b7a583b6f50fc98d2b5e</sub>

packages/babylon-lite/src/text/text-data.ts
```diff
@@ -612,34 +627,40 @@ function applyReplaceRun(data: TextData, prevRef: GlyphRun | number, newRun: Gly
...
+    // the remove path knows how to retire a group.
+    if (newRun.curveSet === group.curveSetId && newRun.glyphs.length > 0) {
+        const prevSlotCount = rec.slots.length;
+        let slots = rec.slots;
+        if (newRun.glyphs.length !== prevSlotCount) {
+            // Glyph count changed, so this run hands its slots back to the group's free list and
+            // takes a fresh block. It reclaims most of them immediately — the allocator pops the
+            // slots it just freed — so the write stays within roughly the same buffer range.
+            freeSlots(data, group, slots);
+            slots = allocateSlots(data, group, newRun.glyphs.length);
```
deltakosh: Resizing a run reverses its reused slots because `freeSlots` pushes the old slots in glyph order and `allocateSlots` pops them LIFO. For example, resizing slots `[0, 1]` to three glyphs yields `[1, 0, 2]`. Instances are drawn in slot order, so overlapping glyphs (especially with distinct per-glyph colors/alpha) change compositing order and pixels. Please preserve glyph order when reclaiming this run's slots.

deltakosh (commented):
Found one draw-order regression in the new in-place reslot path.

packages/babylon-lite/src/text/text-data.ts
ryantrem: Good catch — confirmed and fixed in be5ac06b.

Reproduced exactly as you described: growing a run from 2 to 3 glyphs gave `[1, 0, 2]`, and a repeated grow/shrink loop produced `[5, 4, 0]`. `freeSlots` pushes in glyph order, `allocateSlots` pops LIFO, so the run came back reversed.

Rather than free-and-reallocate the whole block, the resize path now resizes the run's own slot list in place:
- **shrinking** — hand back only the tail, so surviving slots keep their order
- **growing** — append the newly allocated slots to the ones the run already holds

Slots reclaimed from the free list can still land *below* the ones the run holds, so the appended range is checked and the list sorted only when that actually happens; growth off the group tail (the common case) skips the sort entirely. This also drops a free-list round trip on every resize, so it is slightly faster than what was there before.

Kept it allocation-free (in-place `length` truncation and `push`, no `slice`/`concat`) so the run's slot array is never reallocated and keeps its `PACKED_SMI_ELEMENTS` kind — that property is load-bearing for `shiftSlotsAtOrAfter`. Verified with `%HasHoleyElements` that truncation, `push`, and `sort` all preserve it.

Added regression coverage asserting ascending slot order after grow, after shrink, and across 40 rounds of mixed resizes. All three fail on the previous commit.

Validation: lint clean, 1317/1317 unit tests, 223/223 bundle-size ceilings, scene275 text parity passes. The fix costs +233 B raw on scene180/181 (manifests updated); both stay under their ceilings.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/520/)

<sub>Build 20260803.14 - merge @ fc6c1fbc564e4cd732e2da812c80ee746ca5c791</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260803.14/lab/index.html)

<sub>Build 20260803.14 - merge @ fc6c1fbc564e4cd732e2da812c80ee746ca5c791</sub>

packages/babylon-lite/src/text/text-data.ts
ryantrem: I'm still iterating on this, maybe with a better solution... stand by.

ryantrem: Landed a better fix in a4f1485c — replacing the in-place resize from be5ac06b.

Digging into this further turned up two things. First, the reversal isn't limited to the resize path, and it isn't new to this PR: `applyAddRun` reuses freed slots through the same LIFO pop, so removing a run and adding another already returns `[3, 2]` on master today. Second, resizing in place only patched the one path I'd been shown, and cost 233 raw bytes doing it.

So the guarantee now lives in `allocateSlots`, where every caller picks it up. Order is tracked while the free list drains and the sort fires only when a reclaimed block actually came back reversed. The guard matters: `sort` runs a comparator call per element even on already-ordered input, and an unconditional sort took cross-group `replaceRun` from -4.9% to +1.5% against master. Guarding it recovers that completely.

Two incidental wins fell out of it — nothing refills the free list mid-drain, so the loop now breaks at the first miss instead of calling `popFreeSlot` on an empty list for every remaining glyph, and the extension writes the tail of `out` directly instead of rescanning for `-1` placeholders.

Interleaved A/B against master, 660 runs x 18 glyphs, median of 160 rounds with variant order rotated each round:

| path | vs master |
| --- | --- |
| addRun full doc | -16.2% |
| replaceRun same-size | -22.9% |
| replaceRun resize | -2.7% |
| replaceRun cross-group | -4.7% |

The ordering fix now costs 69 raw bytes over the unguarded branch (scene180 +207, scene181 +193 vs merge-base), down from 233, with no movement in gzip.

Added coverage for the add path reusing freed slots — it reproduces `[3, 2]` without this change. Full unit suite green at 1333, lint and typechecks clean, bundle-size parity 223/223, scene275 text parity passing.

One thing worth flagging: the in-place resize was actually faster on the resize path specifically (-10.4% vs -2.7%), since it skips the free/realloc round trip entirely. I didn't keep it because that path measures ~0.5 ms in real LineLayoutPad usage against ~10 ms for cross-group, so it looked like the wrong 164 bytes to spend. Happy to put it back if you'd rather have it.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/520/)

<sub>Build 20260803.18 - merge @ a801a195e0ec785045167259003a36a1865b43a9</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260803.18/lab/index.html)

<sub>Build 20260803.18 - merge @ a801a195e0ec785045167259003a36a1865b43a9</sub>

deltakosh (approved):

PR #525, opened 2026-08-04
fix!: compose hierarchy instance matrices with the template hierarchy

RaananW:
Fixes the forum report: [glTF thin instances require explicit negative X scaling](https://forum.babylonjs.com/t/gltf-thin-instances-require-explicit-negative-x-scaling/63886) ([repro playground](https://liteplayground.babylonjs.com/snippet/N601SI/v/3)).

## Problem

`createHierarchyInstancePool()` baked `rootWorld⁻¹` into every binding:

```
instanceMatrix = meshWorld⁻¹ · M · rootWorld⁻¹ · meshWorld
finalWorld     = mesh.world · instanceMatrix = M · rootWorld⁻¹ · meshWorld
```

So the instance matrix **replaced** the template root's world matrix rather than composing with it.

A `loadGltf()` hierarchy's `__root__` node carries the RH→LH conversion as scaling `(-1, 1, 1)` (`load-gltf.ts:350`). Dividing it out mirrors every instance and inverts its winding, which is why the reporter had to pass an explicit `(-1, 1, 1)` scale just to get the model back to its authored orientation. A coordinate-system conversion is not user placement and must never be cancelled by an instancing helper.

## Fix

Store `meshWorld` instead of `rootWorld⁻¹ · meshWorld`, so the stored per-mesh matrix is `meshWorld⁻¹ · M · meshWorld` and:

```
finalWorld = M · meshWorld
```

The instance matrix now behaves like a parent transform node above the hierarchy root — identity reproduces the template unchanged, and translations/rotations apply in world space on top of it. This is exactly what the existing unit test already asserted (`expected = rootInstance * mesh.worldMatrix`); it passed only because that test's template root world was identity.

Also drops the now-dead root-inversion path and its `"requires an invertible root world matrix"` error.

## Breaking change

`addHierarchyInstance()` / `setHierarchyInstanceMatrix()` matrices compose with the template hierarchy instead of replacing the root world matrix. Only affects templates whose root world is not identity (every glTF). Callers relying on the old semantics should pass `matrix * rootWorld⁻¹`.

## Validation

- New regression test with a flipped glTF-style root (`scaling (-1, 1, 1)`): identity instance is a no-op, and a translated instance equals `offset * templateWorld`. Verified it **fails** on `master` and passes with the fix.
- `vitest run tests/lite/unit/hierarchy-instance-pool.test.ts` — 5 passed.
- `tsc -p packages/babylon-lite/tsconfig.json --noEmit` and `tsc -p tests/lite/tsconfig.json --noEmit` clean; eslint clean on changed files.
- No scene or lab code uses hierarchy instance pools, so no parity/bundle-manifest impact (the change is net code removal).


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/525/)

<sub>Build 20260804.3 - merge @ 4b1c7678935414f603f410cbab8b542d48618a79</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/525/)

<sub>Build 20260804.6 - merge @ 8cc48cc31e0e297e200cade657ca5025f01e6dff</sub>

deltakosh (approved):

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/525/)

<sub>Build 20260804.10 - merge @ c07e0f4800b62be72214bea151ea915c19e98a5e</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/525/)

<sub>Build 20260804.13 - merge @ 078a6ec84810fb00331ffc1a44c11f8d7a484aa5</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260804.13/lab/index.html)

<sub>Build 20260804.13 - merge @ 078a6ec84810fb00331ffc1a44c11f8d7a484aa5</sub>

PR #555, opened 2026-08-13
perf(text): specialize one Slug fragment shader for A2C instead of duplicating it

ryantrem:
## What

`slug-a2c.frag.wgsl` was a verbatim copy of `slug.frag.wgsl` differing by exactly one line:

```wgsl
// blended  (premultiplied alpha -- the blender does src + dst*(1-srcA))
return in.vColor * coverage;

// A2C      (straight alpha -- the hardware derives the sample mask from alpha)
return vec4<f32>(in.vColor.rgb, in.vColor.a * coverage);
```

The difference is real -- premultiplying under alpha-to-coverage would apply coverage twice, giving edges that are too dark and too thin -- but it is only *whether RGB is coverage-weighted*, which is a boolean.

This PR expresses that boolean as a WebGPU pipeline-overridable constant and deletes the copy:

```wgsl
override a2c: bool = false;
...
let rgbWeight = select(coverage, 1.0, a2c);
return vec4<f32>(in.vColor.rgb * rgbWeight, in.vColor.a * coverage);
```

The A2C pipeline sets `constants: { a2c: 1 }`; the blended pipeline's descriptor is byte-identical to before, and both variants now share a single `GPUShaderModule`. `pipelineKey` already encoded the A2C bit, so pipeline caching is unaffected.

## Why

The duplicate cost **every consumer 3,040 raw bytes** of inlined WGSL whether or not they enabled A2C.

That is invisible to our own bundle gate, and the reason is interesting enough to be worth writing down. The design in `alpha-to-coverage-hook.ts` is a null-resolver seam: `let _resolver = null`, assigned only from the opt-in `alpha-to-coverage.ts`. Rollup proves `_resolver` is always null, folds `alphaToCoverage` to `false`, kills the ternary, and drops the shader string -- `scene180.js` contains **zero** occurrences of `text-a2c-frag`. So the scene manifests correctly showed ~no growth when A2C landed (#450 moved scene180 by +6 bytes).

But that property holds for **exactly one bundler**. A webpack consumer does drop the opt-in module (`setAlphaToCoverage`, `_enabledTargets`, `_isAlphaToCoverageEnabled` are all absent) yet will not do cross-module value propagation, so it retains the branch and the 3 KB string. Closure Compiler in ADVANCED mode does not fold it either -- verified empirically, not assumed. A downstream webpack app that only uses text and sprites was shipping the A2C shader it never enables.

## Honest accounting of the size win

Near-identical strings compress well, so I do not want to oversell this:

| | duplicate shader cost |
|---|---|
| raw | **3,040 B** |
| gzip | ~170 B |
| brotli | ~95 B |

**Over the wire this is worth ~95 bytes, not 3 KB.** The genuine wins are raw-byte budget headroom, parse/eval and memory cost of 3 KB less JS string, and removing a duplicated source of truth. If you would rather keep two readable shader files and accept the bytes, that is a defensible call -- I would just want the tradeoff to be explicit rather than accidental.

### Measured impact

| scene | raw delta |
|---|---|
| scene275 (A2C text) | **-3,016 B** |
| scene180 / scene181 (text, no A2C) | +74 B each |

The +74 B is the `override` declaration plus the `select` -- text scenes that never enable A2C pay a small amount so that A2C consumers stop paying a large amount. A full 237-scene rebuild showed **no other scene moving more than 10 bytes** (sub-tolerance minifier allocation noise), so those three manifests are the only ones committed here.

scene275's ceiling is ratcheted **42.5 -> 39.5 KB** to lock the win in.

## Test change

The unit test that asserted the two shader files had not drifted is now impossible -- and was only ever a workaround for having two sources of truth. It is replaced by one asserting the single-module design directly: both pipelines share `fragment.module`, only the A2C pipeline sets `constants.a2c === 1`, and exactly one fragment module is ever compiled.

## Validation

- **scene275 parity on a real GPU: MAD 0.040 (limit 0.1)** -- this is the important one. It exercises the `constants: { a2c: 1 }` specialization path end to end, which the mock-device unit tests structurally cannot.
- scene274 parity passes.
- 1,546 unit tests pass (full suite, post-merge with master).
- 233 bundle-size ceiling tests pass.
- Full `build:bundle-scenes` (237 scenes) -- manifest validator flags exactly the three scenes above.
- eslint / prettier / tsc clean.

## Note on the WGSL minifier

I checked before relying on this: `scripts/wgsl-minify-plugin.ts` applies its `mangle` option only in `renderChunk`, so `?raw` `.wgsl` files always go through `minifyWgslMiniray(src, {})` with defaults and *are* mangled and tree-shaken. I probed miniray directly to confirm it preserves `override a2c` by name and does not fold `select(coverage, 1.0, a2c)` -- it treats override names as pipeline interface, like bindings. No plugin change was needed. The short name is deliberate and matches existing precedent (`rgbd-decode.ts` uses `override f`), since the name survives verbatim into every build.

## Possible follow-up (not in this PR)

Our bundle gate is Rollup-only, so it structurally cannot see this class of regression -- any "tree-shaken by a null seam" claim is really "tree-shaken by Rollup." If that pattern is load-bearing elsewhere, adding one webpack or Closure scene to the gate would catch it. Happy to file an issue if you think it is worth it.


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/555/)

<sub>Build 20260813.12 - merge @ 7c43f13e364887d0a50c84e581444972e7dba183</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 275 — MSDF Text Alpha-to-Coverage<br/>`scene275` | 39 KB | 42 KB | -3 KB |

tests/lite/unit/alpha-to-coverage.test.ts
```diff
@@ -257,10 +260,26 @@ describe("WebGPU alpha-to-coverage", () => {
...
+        const blended = getOrCreateTextPipeline(engine, "rgba8unorm", 1, "depth24plus", true, text).pipeline;
+
+        expect(fragmentStage(enabled).module).toBe(fragmentStage(blended).module);
+        expect(fragmentStage(enabled).constants).toEqual({ a2c: 1 });
+        expect(fragmentStage(blended).constants).toBeUndefined();
+
+        // Exactly one fragment module is ever compiled, and it declares the override.
+        const fragmentSources = createShaderModule.mock.calls.map((call) => call[0].code).filter((code) => code.includes("@location(0) vec4<f32>"));
+        expect(fragmentSources).toHaveLength(1);
+        expect(blendedTextFragment).toMatch(/override\s+a2c\s*:\s*bool/);
```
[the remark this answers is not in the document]
ryantrem: Good catch — fixed in `1b550c04`.

You're right that the assertion didn't prove what its comment claimed. It read `override a2c` out of the imported `.wgsl` file text, so a pipeline that compiled some *other* fragment source would still have passed while the compiled module lacked the override. Now asserting against the captured `createShaderModule` source:

```ts
expect(fragmentSources[0]).toMatch(/override\s+a2c\s*:\s*bool/);
```

The imported-file binding is gone entirely, so the test can't drift back.

Worth noting this is stronger than just fixing the indirection: the `.wgsl` files go through miniray, which mangles identifiers. It preserves `override` names because it treats them as pipeline interface (like bindings) — but that's a property of the minifier, not a guarantee. If it ever changed, `constants: { a2c: 1 }` would fail at pipeline creation *at runtime*. Asserting on the compiled source turns that into a unit-test failure instead.

Verified: 10/10 in the suite, eslint/prettier/tsc clean.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/555/)

<sub>Build 20260813.15 - merge @ fe01f06a72f22bbffbdf9dfc3121ea7c22eb1b4f</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 275 — MSDF Text Alpha-to-Coverage<br/>`scene275` | 39 KB | 42 KB | -3 KB |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260813.15/lab/index.html)

<sub>Build 20260813.15 - merge @ fe01f06a72f22bbffbdf9dfc3121ea7c22eb1b4f</sub>

packages/babylon-lite/src/text/_gpu/text-pipeline.ts
```diff
@@ -106,8 +103,12 @@ export function getOrCreateTextPipeline(
             ],
         },
         fragment: {
-            module: fragModule,
+            module: cache.fragModule,
             entryPoint: "main",
+            // `a2c` is a pipeline-overridable constant in slug.frag.wgsl; setting it switches the
+            // fragment to straight-alpha output. Specialising one module beats shipping a second
+            // near-identical shader, whose text every consumer would pay for even unused.
+            ...(alphaToCoverage ? { constants: { a2c: 1 } } : {}),
```
VicenteCartas: **Blocking:** Closure ADVANCED property-mangles this unquoted WGSL override key. A minimal compile of `globalThis["shader"]="override a2c:bool=false"; globalThis["descriptor"]={"constants":{a2c:1}};` produces `...descriptor={constants:{g:1}}` while the WGSL string remains `override a2c`. WebGPU then receives an unknown constant key and pipeline creation fails for A2C consumers built with Closure, which is one of the downstream compilers this PR targets.

Please make the interface key minifier-proof. The most robust form is a numeric pipeline constant ID: `@id(0) override a2c: bool = false;` paired with `constants: { 0: 1 }` (quoting `"a2c"` would also prevent property mangling).

ryantrem: Good catch — confirmed and fixed in 396ccfda.

You're right on both the mechanism and the blast radius. `a2c` was an unquoted object-literal key with no externs entry, and the WGSL is an opaque string to the compiler, so Closure ADVANCED renames one side and not the other. The failure is also silent at build time and only shows up as a pipeline-creation error on A2C consumers, which makes it worse than a plain size bug.

Took the numeric-ID form you recommended rather than quoting:

- `slug.frag.wgsl`: `@id(0) override a2c: bool = false;`
- `text-pipeline.ts`: `constants: { [A2C_CONSTANT_ID]: 1 }` with `A2C_CONSTANT_ID = 0`

Numeric keys can't be property-mangled by any JS minifier, and per the WebGPU spec the numeric key is actually *required* once `@id` is specified, so this is the canonical form rather than a workaround. Quoting `"a2c"` would have worked too, but it stays correct only as long as nobody writes it unquoted later; the numeric ID removes the footgun instead of documenting it. Kept the `a2c` name in the WGSL for readability and left a comment at both sites explaining why the ID is explicit, since `@id(0)` is otherwise non-obvious.

The test now pins the invariant instead of the value — it asserts the constants key set is exactly `["0"]` and that the compiled source declares `@id(0) override a2c`, so a revert to a name-keyed constant fails the suite rather than silently shipping.

Cost: scene275 39,976 -> 39,989 B raw, still well under the 39.5 KB ceiling (the added comments are stripped by the WGSL minifier, so it's just the attribute). Manifests regenerated, all 237 validate, 1546 unit tests green.

One thing your comment surfaced that's outside this PR: the same name-keyed pattern already ships in two other places, both with an `override f:bool=false` compute shader keyed as `constants: { f: ... }`:

- `packages/babylon-lite/src/loader-env/rgbd-decode.ts:26`
- `packages/babylon-lite/src/loader-gltf/ibl-cubemap-upload.ts:33`

Those have the identical Closure exposure and are already on master. I've deliberately left them out to keep this PR scoped to the regression it introduced — happy to fix them here if you'd prefer, or file a follow-up.

VicenteCartas (approved):

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/555/)

<sub>Build 20260813.17 - merge @ 05142bcde0328287a215811e2c6e99ae6ac564b9</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 275 — MSDF Text Alpha-to-Coverage<br/>`scene275` | 39 KB | 42 KB | -3 KB |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260813.17/lab/index.html)

<sub>Build 20260813.17 - merge @ 05142bcde0328287a215811e2c6e99ae6ac564b9</sub>

PR #573, opened 2026-08-16
fix(math): mat4 factories must allocate through allocateMat4

VoxelBoy:
Eight mat4 factories allocate with `new F32(16)` instead of `allocateMat4()`, so the matrices they return stay F32 however the engine was configured:

```
mat4-compose.ts   mat4-from-quat.ts       mat4-identity.ts     mat4-invert.ts
mat4-look-at-lh.ts  mat4-multiply.ts      mat4-perspective-lh.ts  mat4-scale.ts
```

> Originally opened against seven; `mat4-look-at-lh.ts` was found in review (thanks @deltakosh) and is fixed in the same branch. Counts throughout this description have been corrected to eight.

This is load-bearing for **root** nodes. `createWorldMatrixState` allocates `_ownedWorld` through the allocator but uses it only when a node has a parent; for a root, `getWorldMatrix()` returns the local matrix unchanged, and that came from `composeTrsLocalMatrix` → `mat4Compose`/`mat4Identity`. So under `useHighPrecisionMatrix: true` every root-level object silently kept an F32 world transform, while parented nodes looked correct and masked it. At 4.6e6 m the F32 ULP is ~0.5 m: six consecutive 1 cm nudges of a root node's position all produced an identical `worldMatrix[12]`.

Both `21-core-math.md` ("New matrices are allocated via `allocateMat4()`…") and `36-high-precision-matrix.md` ("every matrix allocation on the page returns `Float64Array(16)`") already document the contract this restores, so this is code-vs-doc divergence rather than a design change. The second commit adds an allocation-boundary inventory to `36-high-precision-matrix.md`, alongside the GPU-upload inventory that already existed — the missing half is how these eight drifted.

The cast in each change is needed because `allocateMat4()` returns the read-only `Mat4`, which these files already distinguish from the writable `Mat4Storage`.

## Tests

`tests/lite/unit/engine-matrix-policy.test.ts` — per-factory assertions (so a regression names the factory), a behavioural check that a root node keeps sub-metre precision at 4.6e6 under HPM, and an F32-default control without which the behavioural test would pass with or without the fix. All three verified failing before the change.

## Checks

- `vitest --project unit` — 1576 passed / 2 skipped, 205 files
- Bundle size: a net reduction across the fleet. `allocateMat4()` minifies smaller than an inlined `new F32(16)`, and the allocator module is already in every bundle. A full `build:bundle-scenes` measures **221 scenes changed, −2801 B net, −35 B at best**; two NPE scenes move the other way by +6 B and +1 B (scene277, scene281), both with KBs of headroom. No ceiling moves and `scene-config.json` is untouched. The regenerated `lab/public/bundle/manifest/` files are committed — omitting them is what failed the Bundle Size job on the first push, since `validate:bundle-manifest` trips once committed values drift past its 10 B tolerance.
- Per `AGENTS.md`, the all-scene parity suite was not run locally.

---

Found while porting [Manywhere](https://forum.babylonjs.com/t/lite-no-supported-path-for-gpu-generated-geometry-proposal-prototype/63959), a WebGPU procedural space game, onto Lite. Happy to adjust anything here to fit your conventions.


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260817.1/testResults/parity/index.html)

<sub>Build 20260817.1 · merge @ 20176db444d3be5d5590dca47a59651eaf2bfaf4</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/573/)

<sub>Build 20260817.1 - merge @ 20176db444d3be5d5590dca47a59651eaf2bfaf4</sub>

deltakosh (changes requested):
Nice fix!

The root-vs-parented masking analysis is spot on, and the 4.6e6 m behavioural test (plus the F32 guard-the-guard) is exactly right.

One thing before merge: the invariant isn't fully applied. mat4LookAtLH is an eighth mat4-returning factory that still hand-builds new F32([...]):

// packages/babylon-lite/src/math/mat4-look-at-lh.ts
return new F32([ xAxis.x, yAxis.x, zAxis.x, 0, ... ]) as unknown as Mat4;

It's publicly exported, so under HPM it returns an F32 view matrix: the same loss this PR fixes elsewhere. It's also inconsistent: the return mat4Identity() fallback is now F64 while the normal path is F32. Consequently the guard test's factories map omits it, and the doc says "seven factories" when it's eight.

Suggested: route it through the allocator with an *Into fill, add it to the test map, bump the doc. If it's deliberately exempt, a one-line note would settle it , but since it's public I'd just fix it.

Nit: allocateMat4() as unknown as Mat4Storage now repeats at all seven sites; a small allocateMat4Storage() helper would centralize it.

Otherwise LGTM — happy to approve once look-at is in.

VoxelBoy:
You were right, and it was worse than one factory: `mat4LookAtLH` is fixed, but chasing the red checks turned up a stale-artifact problem that had nothing to do with the review.

**Look-at.** Converted to the same allocate + `*ToRef` shape as its sibling `mat4LookAtWorldLHToRef` — a new `mat4-look-at-lh-to-ref.ts` writes into caller storage, `mat4LookAtLH` allocates and delegates. I used `*ToRef` rather than the `*Into` you suggested because the module already splits those two ways: the view/projection builders are `mat4LookAtWorldLHToRef`, `mat4PerspectiveLHToRef`, `mat4OrthoOffCenterLHToRef`, while `*Into` is the composition family (`mat4ComposeInto`, `mat4MultiplyInto`, `mat4FromQuatInto`). Look-at sits with the first group. Happy to rename if you read that split differently. The degenerate-input fallback now writes identity in place, so both paths finally agree in storage type. Added to the guard test's factories map (verified the assertion fails without the fix), and the doc says eight everywhere.

**The helper — I did not do what you asked, and I want to flag that rather than bury it.** `allocateMat4Storage()` exists, but only at the new look-at site. Routing the seven pre-existing factories through it costs real bytes:

| | scene1 | scene7 | scene14 |
| --- | --- | --- | --- |
| as-is (inline cast) | 89869 B | 107753 B | 90131 B |
| helper at all 8 sites | +39 | +35 | +36 |
| helper at look-at only | 0 | 0 | 0 |

The cast is erased at compile time; a real wrapper function isn't, and the minifier doesn't inline it back out. It lands on every scene that composes a transform, which is nearly all of them, and scene7 had 141–176 B of headroom. `mat4LookAtLH` has no current scene traffic, so it's free there. Happy to take the ~35 B fleet-wide if you'd rather have the consistency — it's your call, I just didn't want to spend it silently.

**The two red checks were not what they looked like.** Bundle Size: the original fix is a genuine *reduction*, but the PR never regenerated `lab/public/bundle/manifest/`, and `validate:bundle-manifest` fails once committed values drift past its 10 B tolerance. That was true of **221 scenes**, not the handful I'd assumed — net −2801 B across the fleet, best case −35 B on a single scene. Regenerated with a full `build:bundle-scenes` and committed (222 manifest files; one moved only its chunk hashes); validation now reports 237 scenes up to date. Two NPE scenes move the other way (+6 B scene277, +1 B scene281), both with KBs of headroom, so no ceiling moves and `scene-config.json` is untouched. 113/114/115 left at their CI-authored values.

API Report I could not reproduce: running api-extractor against the built `.d.ts` with the same config gives 0 errors, and the generated `.api.md` is byte-identical to a baseline built at the merge-base. Neither new symbol is exported from the package entry. I don't have a fix because I can't find a fault — flagging it as unexplained rather than claiming otherwise.

**One thing you may want to look at beyond this PR.** Comparing job durations across my four PRs, the failures split into two kinds. #574's Bundle Size failure ran 27m50s and found a real ceiling breach. But #573's API Report failed in **9s** and Bundle Size in **18s**, and #575's Unit Tests and Parity both failed in **32s** while the run reported 802 passed / 0 failed. Sub-minute failures on jobs whose passing runs take 3–40 minutes look like aborts before the work starts, not results. If that's a known flake you can ignore this; if not, it may be worth a look, since a red check that isn't a finding is expensive for everyone reading it.

**Merge order.** Those regenerated manifests include all 14 that #574 touches, so despite what both descriptions say, these two are no longer independent — whichever lands second will need its manifests regenerated. Happy to rebase whichever you'd rather take last.

Checks: `tsc --noEmit` clean on both projects, focused unit tests green (74/74 across the matrix/allocator/floating-origin files), `validate:bundle-manifest` clean, full `build:bundle-scenes` run locally for the manifests.

VoxelBoy:
One correction to the merge-order note above: it named only #573 and #574, because #576 had no manifests when I wrote it. It does now — the same missing-manifest defect was latent there and is fixed in `de96ca08`. All three overlap: #573 ∩ #576 = 124 files, #573 ∩ #574 = 14, #574 ∩ #576 = 7. The full table and reasoning are in https://github.com/BabylonJS/Babylon-Lite/pull/576#issuecomment-5321719538 — no action needed here, just so the picture is accurate before you pick a merge order.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/573/)

<sub>Build 20260818.10 - merge @ 9f5c48fddb8a7a38290b607f13aac3679708b6dd</sub>

deltakosh (approved):

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 98 — Billboard Additive Blend<br/>`scene98` | 58 KB | 59 KB | -1 KB |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/573/)

<sub>Build 20260819.4 - merge @ 2ff226eed3a5894fe7cad3235aa1a14516cb97b2</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 98 — Billboard Additive Blend<br/>`scene98` | 58 KB | 59 KB | -1 KB |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/573/)

<sub>Build 20260819.5 - merge @ 94bae461e67a640f00a5397f4de1944f2f50b4a4</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 98 — Billboard Additive Blend<br/>`scene98` | 58 KB | 59 KB | -1 KB |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260819.5/lab/index.html)

<sub>Build 20260819.5 - merge @ 94bae461e67a640f00a5397f4de1944f2f50b4a4</sub>

PR #574, opened 2026-08-16
fix(shader)!: ShaderMaterial must honour the floating origin

VoxelBoy:
## The bug

`standard`, `pbr` and `node` renderables rebase a mesh's world matrix onto the camera before upload, resolving a packer once at construction:

```js
const _packMeshWorld = engine._makePackMeshWorld?.(scene) ?? packMat4IntoF32;
_packMeshWorld(meshUboData, mesh.worldMatrix, 0, 0);
```

Both `ShaderMaterial` system-uniform writers read `mesh.worldMatrix` raw instead — `shader-renderable.ts` → `writeSystemUniforms`, and `enable-shader-material-uniform-caching.ts` → `writeCachedSystemUniforms`.

Under `useFloatingOrigin` this is not a precision detail, it is a wrong picture. `getViewMatrix` forces the view translation to zero because it expects the mesh pack to have established the eye-relative frame, so an un-rebased mesh draws **as though the camera sat at the world origin**. Its apparent position and size then depend only on where it is, never on where the viewer is. `LineMaterial` is a `ShaderMaterial` underneath, so line meshes are affected too.

## Repro

A mesh at world x = 4,600,000 seen by a floating-origin camera 4,561,405 m away:

- **Expected:** `world[12] === 4561405` (camera-relative)
- **Actual:** `world[12] === 4600000` (absolute)

Consequently `worldViewProjection` is *identical* at 4,561 km and at 304 km. Flying toward a 75,977 m-radius planet, the distance readout counted down from 4,561 km to 304 km and terrain LOD subdivided correctly — 41,472 → 131,328 visible triangles — while the disc, which should have grown from 1.9° to 28°, did not move a pixel. That split is the diagnostic signature: the application knew the distance, the renderer did not.

## The fix

`_shaderWorldMatrix` does the subtraction both writers share. It cannot call `packMat4IntoF32WithOffset` — that lives in the LWR-only bundle this module must not statically import.

- The rebased matrix flows into **`worldView` and `worldViewProjection`** as well as `world`. They all derive from it; fixing only one would tear geometry apart between passes reading different uniforms.
- **`cameraPosition` now reports zero under FO.** See the behaviour note below.
- Keyed on **`camera._useFloatingOrigin`**, not the engine flag, so the test is bit-for-bit the one `getViewMatrix` applies to the same camera. A render-target task drawing through a non-scene camera then gets an untranslated view *and* an absolute world, which is self-consistent.

Returns `mesh.worldMatrix` untouched when FO is off, so the non-LWR path stays copy-free.

## Behaviour change: `cameraPosition` under floating origin

Custom shaders that read the `cameraPosition` system uniform will see `(0, 0, 0)` under `useFloatingOrigin` instead of the absolute eye position.

I've treated this as part of the fix rather than a separate breaking change, because `24-shader-material.md` already documents `cameraPosition` as equivalent to `scene.vEyePosition.xyz`, and `_packSceneUniforms` already zeroes `vEyePosition` under FO. The current behaviour is what diverges from the documented contract. It is also the only self-consistent choice once `world` is camera-relative — an absolute eye alongside camera-relative geometry is two coordinate frames at once, and the common `cameraPosition - worldPos` expression keeps working (now at full precision) only if both sides move together.

Flagging it explicitly so you can decide whether it warrants a `breaking` label and the release marker `AGENTS.md` requires; happy to add one if you'd rather treat it that way.

## Bundle size — needs a decision

**This is the part I can't resolve from outside.** The fix adds ~230–240 B to every scene that reaches a ShaderMaterial, and most of those sit within ~250 B of their ceiling.

> **Corrected after a full-fleet rebuild.** This section originally listed eight scenes with six over ceiling. The real figures, measured by `build:bundle-scenes` across all 237 scenes rather than a grep-derived shortlist: **14 scenes change, 7 exceed.** The additions are `scene213` (GridMaterial is built on `createShaderMaterial` — it exceeds, and was missing entirely), plus `scene127`/`scene128` (via `createLinearDepthMaterial`) and `scene221`/`scene222`, none of which can exceed. Ceilings raised for the seven, per @deltakosh's approval below.

| Scene | Delta | Over ceiling |
| --- | --- | --- |
| scene161 | +242 B | yes |
| scene160 | +240 B | yes |
| scene163 | +239 B | yes |
| scene165 | +239 B | yes |
| scene159 | +236 B | yes |
| scene213 | +231 B | yes |
| scene162 | +230 B | yes |
| scene274 | +242 B | no — 218 B headroom |
| scene128 | +239 B | no — no `maxRawKB` |
| scene222 | +239 B | no — 5.4 KB headroom |
| scene221 | +237 B | no — 3.6 KB headroom |
| scene278 | +236 B | no |
| scene127 | +235 B | no — no `maxRawKB` |
| scene279 | +234 B | no |

Scenes that never reach a ShaderMaterial are unaffected — scene2, scene200 and scene205 are byte-identical to master, and `scene49` measures exactly 0 despite matching the same source grep.

The added code cannot be tree-shaken: the floating-origin test is a runtime property read on the camera, so the branch ships even in scenes that never enable FO. I looked at moving the subtraction behind an LWR-installed hook (the `engine._makePackMeshWorld` idiom) so the body would live only in FO bundles; that cuts the cost to roughly 50 B but still exceeds five of the six ceilings, so it buys complexity without solving the problem.

Per `AGENTS.md` I have **not** touched `scene-config.json` — ceilings need your explicit approval, and I'd rather not distort the code to chase the threshold. Two paths, your call:

1. Raise the six ceilings by ~250 B each.
2. You'd prefer the LWR-hook indirection anyway on architectural grounds, and raise the ceilings by ~60 B instead. Happy to implement it.

## Tests

`tests/lite/unit/shader-material-floating-origin.test.ts` asserts **approach**, not pack format — move the camera toward a fixed mesh and require what you would see to change:

- projected `w` (view-space depth) shrinks in proportion to the real distance
- the on-screen span of a point one radius off-axis grows in the same proportion
- `world` / `worldView` / `worldViewProjection` stay in one frame
- the eye sits at the origin under FO, and at its real position without it
- sub-metre offsets survive that an F32 absolute world matrix would swallow
- **FO-off controls**, without which a fix that simply zeroed the translation would pass

Six of the seven fail against the current writers; the survivors are the controls. One case in `shader-material-uniform-caching.test.ts` covers the cached writer. That file's mock now spreads `importOriginal` — a mock that dropped the shared helper would have asserted against a stub.

`_defaultShaderSystemUniformWriter` is exported so the test drives the writer that actually ships rather than a stand-in; measured at **zero** bundle cost (builds are byte-identical with and without it), since nothing the package entry reaches refers to it.

## Not a bug, checked while here

`_worldCenter` stays absolute for transparency sorting in every renderable family. Under FO the view matrix has no translation, so the omitted term is constant across all renderables in a frame and cancels in the comparison. Left alone.

## Checks

- `vitest --project unit` — 1581 passed / 2 skipped, 206 files
- Bundle size — measured, see above
- Per `AGENTS.md`, the all-scene parity suite was not run locally

Second commit adds the docs: ShaderMaterial was missing from the LWR doc's "three places the offset is subtracted" and fell outside the HPM doc's "exhaustive" GPU-upload boundary list, which is the same gap the code had.

---

Found while porting [Manywhere](https://forum.babylonjs.com/t/lite-no-supported-path-for-gpu-generated-geometry-proposal-prototype/63959), a WebGPU procedural space game, onto Lite. Happy to adjust anything here to fit your conventions.


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/574/)

<sub>Build 20260817.2 - merge @ f3c64c3e4a83d341f9be35080b2766cb1d018d6d</sub>

deltakosh:
Really clean fix, and the "the game knew the distance, the renderer did not" diagnosis nails it. 

Two calls for you:

- Bundle ceilings. I'd take option 1 (raise the six ceilings ~250 B) and skip the LWR-hook indirection, since the hook still busts five ceilings. Not worth distorting the code for ~60 B.
- cameraPosition to (0,0,0) under FO is user-visible. It matches the documented vEyePosition contract so it belongs here, but it needs the breaking label and release marker.

Two nits:

* _shaderWorldMatrix returns the shared _foWorldScratch by reference. Safe today, but since it's exported and tested directly, an optional out-param removes the aliasing footgun without losing the copy-free path.
* Test parity. The suite proves ShaderMaterial is self-consistent but not that it matches the packMat4IntoF32WithOffset path for the same mesh, which is the exact divergence that caused this. One assertion tying them together keeps them locked as either side evolves.

Otherwise LGTM, happy to approve once the ceilings and breaking label are sorted.

deltakosh (changes requested):
Two calls for you:

- Bundle ceilings. I'd take option 1 (raise the six ceilings ~250 B) and skip the LWR-hook indirection, since the hook still busts five ceilings. Not worth distorting the code for ~60 B.
- cameraPosition to (0,0,0) under FO is user-visible. It matches the documented vEyePosition contract so it belongs here, but it needs the breaking label and release marker.

Two nits:

* _shaderWorldMatrix returns the shared _foWorldScratch by reference. Safe today, but since it's exported and tested directly, an optional out-param removes the aliasing footgun without losing the copy-free path.
* Test parity. The suite proves ShaderMaterial is self-consistent but not that it matches the packMat4IntoF32WithOffset path for the same mesh, which is the exact divergence that caused this. One assertion tying them together keeps them locked as either side evolves.

bjsplat:
## Bundle Size Changes

### Increases

| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 61 KB | 60 KB | **+1 KB** |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 61 KB | 60 KB | **+1 KB** |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 45 KB | 44 KB | **+1 KB** |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 48 KB | 47 KB | **+1 KB** |

VoxelBoy:
Both calls taken, both nits done — and re-measuring the ceilings turned up a seventh scene I'd missed, so please read the first section before approving.

**Ceilings — it's seven, not six.** I rebuilt the whole fleet rather than trust my original inventory, and `scene213-gridmaterial` breaches too: GridMaterial is built on `createShaderMaterial`, and my eight-scene list in the PR description simply didn't have it. Raised the same ~250 B as the others (49.27 → 49.52 KB). The full list now raised: scene159, 160, 161, 162, 163, 165, 213. A full `build:bundle-scenes` passes with no ceiling exceeded, so I'm now confident this is the complete set rather than the set I happened to grep for.

Two corrections to the table in the description while I'm here:

- **scene127/128 (Gaussian Splatting) do move** — +235 B and +239 B, via `createLinearDepthMaterial`, another ShaderMaterial consumer I'd missed. They have no `maxRawKB` at all, so they can't breach, but the description's claim that non-ShaderMaterial scenes are byte-identical was wrong.
- **scene221 and scene222 grow too** (+237 B, +239 B), but they carry 3.6 KB and 5.4 KB of headroom so nothing moves for them. scene49 is the only candidate that measured exactly 0.

Which leads to something worth flagging about the bundle bot: the four scenes it commented on (127, 128, 165, 279) are exactly the ones whose *rounded* KB total crosses an integer boundary — scene127 goes 60.33 → 60.56 KB and reads as 60 → 61. scene213 moves 49.2 → 49.4 KB, crosses no boundary, is not in the bot's list, and breaches its exact-byte ceiling anyway. The bot's rounding means its comment isn't a reliable breach signal in either direction. Not my call what to do about it, but it's what let me ship a wrong table in the description with a green-looking comment underneath.

**`cameraPosition` / breaking.** Agreed. The commit carries a `BREAKING CHANGE:` trailer describing the migration: under `useFloatingOrigin`, `cameraPosition` reports `(0,0,0)` to match the eye-relative frame `world`/`worldView`/`worldViewProjection` are rebased into; shaders reading it as an absolute position need to derive that from `world` instead; `cameraPosition - worldPos` needs no change and now returns the eye vector at full precision. I've left the `breaking` label and the PR title to you — say the word and I'll retitle to `fix(shader)!:`, or apply it yourself if that's easier.

**Out-param.** `_shaderWorldMatrix(mesh, camera, out?)`. Without `out`, behaviour is unchanged — shared scratch under FO, `mesh.worldMatrix` by reference when FO is off, no allocation. With `out`, both branches write into it, including the FO-off path that previously aliased `mesh.worldMatrix`, so passing `out` always means "the answer is in your buffer." Both call sites still pass nothing. Measured cost: **+2 B**.

**Parity assertion.** Added: same mesh, same FO camera, `_shaderWorldMatrix`'s output asserted byte-identical to `packMat4IntoF32WithOffset`'s — they do the same F64 subtraction and F32 store, so there's no legitimate difference to allow for, and I've noted that inline. Verified it fails against the pre-fix writer, showing `4600000` against the packer's `303913`, which is the original bug's signature exactly.

**One practical note on merge order.** #573 regenerates 222 bundle manifests (it's a fleet-wide byte *reduction* that was never committed), and all 14 of the manifests this PR touches are inside that set. So these two are no longer independent, whatever the descriptions say — whichever lands second will need its manifests regenerated. Happy to rebase whichever you'd prefer to take last.

Checks: `tsc --noEmit` clean on both projects, 15/15 on the two ShaderMaterial test files, eslint clean, full `build:bundle-scenes` green, `validate:bundle-manifest` reports 237 scenes up to date.

VoxelBoy:
One correction to the merge-order note above: it named only #573 and #574, because #576 had no manifests when I wrote it. It does now — the same missing-manifest defect was latent there and is fixed in `de96ca08`. All three overlap: #573 ∩ #576 = 124 files, #573 ∩ #574 = 14, #574 ∩ #576 = 7. The full table and reasoning are in https://github.com/BabylonJS/Babylon-Lite/pull/576#issuecomment-5321719538 — no action needed here, just so the picture is accurate before you pick a merge order.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/574/)

<sub>Build 20260818.9 - merge @ baeb5cdebfe07e2e9b33f8d2d97a10b7dff1399c</sub>

deltakosh (changes requested):
Please address these changes before merge:

- Preserve the breaking-change marker in the squash metadata: add `!` to the PR title or an exact `BREAKING CHANGE:` trailer to the PR body, and apply the `breaking` label. The marker currently exists only in an intermediate commit and will not survive a squash merge.
- Replace the inline `typeof import(...)` in `tests/lite/unit/shader-material-uniform-caching.test.ts:14` with a top-level type import, as required by `GUIDANCE.md`.

bjsplat:
## Bundle Size Changes

### Increases

| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 61 KB | 60 KB | **+1 KB** |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 61 KB | 60 KB | **+1 KB** |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 45 KB | 44 KB | **+1 KB** |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 48 KB | 47 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260818.9/lab/index.html)

<sub>Build 20260818.9 - merge @ baeb5cdebfe07e2e9b33f8d2d97a10b7dff1399c</sub>

deltakosh:
Addressed the remaining requests in 22b14660: the PR title now carries the breaking-change marker, the new \reaking\ label is applied, and the inline \	ypeof import(...)\ was replaced with a top-level type import per GUIDANCE.md. Merged current master, regenerated the two conflicted manifests, and re-ran the 15 focused ShaderMaterial tests plus scoped ESLint; all pass.

deltakosh (approved):
Requested changes are addressed: breaking squash metadata and label are present, the GUIDANCE.md type-import requirement is fixed, master is merged, and focused validation is green.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/574/)

<sub>Build 20260819.6 - merge @ 675b0e6dbcac37cc896d1732b9535c81ad5684bd</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 61 KB | 60 KB | **+1 KB** |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 61 KB | 60 KB | **+1 KB** |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 45 KB | 44 KB | **+1 KB** |
| Scene 261 — Temporal Anti-Aliasing<br/>`scene261` | 53 KB | 52 KB | **+1 KB** |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 48 KB | 47 KB | **+1 KB** |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/574/)

<sub>Build 20260819.8 - merge @ a07403f42650faf7db6fba4eb4efb140cd1a0965</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 61 KB | 60 KB | **+1 KB** |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 61 KB | 60 KB | **+1 KB** |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 45 KB | 44 KB | **+1 KB** |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 48 KB | 47 KB | **+1 KB** |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/574/)

<sub>Build 20260819.9 - merge @ 810daa585142cdd87e3441a22b741c964b296800</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 61 KB | 60 KB | **+1 KB** |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 61 KB | 60 KB | **+1 KB** |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 45 KB | 44 KB | **+1 KB** |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 48 KB | 47 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260819.9/lab/index.html)

<sub>Build 20260819.9 - merge @ 810daa585142cdd87e3441a22b741c964b296800</sub>

PR #575, opened 2026-08-16
feat(particle): export createParticleSystem from the package entry

VoxelBoy:
`particle/particle-system.js` exports four functions; the barrel re-exports three:

```ts
// packages/babylon-lite/src/index.ts
export { animateParticleSystem, startParticleSystem, stopParticleSystem } from "./particle/particle-system.js";
//       ^ missing: createParticleSystem — the constructor the other three take as their argument
```

The `ParticleSystem` type is exported, and the rendering half (`createParticleBillboard`, `syncParticleBillboard`) is public, so the imperative particle path is reachable in every respect except the call that begins it. One name added to an existing re-export line; no implementation change.

Found while porting a WebGPU procedural space game onto Lite — we had deferred two emitters after reading `index.d.ts` and concluding Lite offered only Node Particle graphs.

### DIFF
--- a/packages/babylon-lite/src/index.ts
+++ b/packages/babylon-lite/src/index.ts
@@ -730,7 +730,7 @@ export { buildNodeParticleSetWithFlowMaps } from "./particle/node/npe-flow-map.j
 export { buildNodeParticleSetWithNoiseTextures } from "./particle/node/npe-noise.js";
 export { buildNodeParticleSetWithTextureUpdates } from "./particle/node/npe-texture-updates.js";
 export type { ParticleSystem } from "./particle/particle-system.js";
-export { animateParticleSystem, startParticleSystem, stopParticleSystem } from "./particle/particle-system.js";
+export { animateParticleSystem, createParticleSystem, startParticleSystem, stopParticleSystem } from "./particle/particle-system.js";
 export type { RegisterNodeParticleOptions } from "./particle/particle-scene.js";
 export { registerNodeParticleSet } from "./particle/particle-scene.js";
 export { createParticleBillboard, syncParticleBillboard } from "./particle/particle-billboard.js";


bjsplat:
### 📋 parity — Test Report

[View full Playwright report](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260817.3/testResults/parity/index.html)

<sub>Build 20260817.3 · merge @ fa9783cf17a3c01f6978654e915eb444dc8a7120</sub>

bjsplat:
## API Changes

<details>

```diff
diff --git a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
index eff3a836..2db448e0 100644
--- a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md
+++ b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
@@ -1514,6 +1514,9 @@ export function createNullEngine(_options?: NullEngineOptions): EngineContext;
 // @public
 export function createParticleBillboard(system: ParticleSystem): FacingBillboardSpriteSystem;
 
+// @public
+export function createParticleSystem(capacity: number): ParticleSystem;
+
 // @public
 export function createPbrMaterial(props?: Partial<PbrMaterialProps>): PbrMaterialProps;
```

</details>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/575/)

<sub>Build 20260817.3 - merge @ fa9783cf17a3c01f6978654e915eb444dc8a7120</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260817.3/lab/index.html)

<sub>Build 20260817.3 - merge @ fa9783cf17a3c01f6978654e915eb444dc8a7120</sub>

ryantrem (approved):

VicenteCartas (changes requested):
The lifecycle functions are intentionally public because buildNodeParticleSet returns per-system runtime handles that users may start, stop, prewarm, or step manually. registerNodeParticleSet handles the common live-playback case automatically. This is a control surface for NPE-built systems, not a general particle-authoring surface. createParticleSystem is intentionally internal because exposing it crosses that boundary.

42-node-particle.md explains this more:

Image

The current goal for lite is only to support NPE based particle systems (still a work in progress), not the traditional non-NPE API that exists in Babylon.js.

But I would like to understand more of your scenario if you don't mind sharing. What is missing from NPE to support your game?

VoxelBoy:
Fair — and §1 is more explicit than I gave it credit for: "it does not expose a Babylon-style particle constructor" is a design requirement, not a description of the current state. I opened this from the barrel's shape rather than the doc's intent, and the PR reads differently once you've read the latter. Treat the export as withdrawn unless you decide otherwise; it isn't the useful part of this.

The useful part is your question. What's missing turns out to be more specific than we thought. NPE has the capability — two implementation details keep us off it.

`ParticleInputBlock` with `contextualValue: 0x18` (`LocalPositionUpdated`, routed to `particle-input-local-block.js`) integrates a particle's position in the emitter's local frame on *every* simulation step, not just at birth — wired through `UpdatePositionBlock` under `SystemBlock.isLocal = true`, that's exactly `isLocal` semantics: local velocity accumulated, reprojected through `emitterWorldMatrix` each call.

I'd rather give you both blockers than have you fix one and find it didn't help.

**The emitter frame can't move.** `emitterWorldMatrix` is consumed once, as a `BuildNodeParticleOptions`/`ParseNodeParticleOptions` field, when the graph is built — and the public `Mat4` is `{ readonly length: 16; readonly [index: number]: number }`, so there's no supported way to refresh it for a later frame. Fine for a torch or a campfire; not enough for one riding a moving body. Our hull travels up to 3.2 km/s against a particle lifetime of at most 0.13 s, so an emitter frozen at build time is a few hundred metres from the ship within one particle's life and the plume detaches.

**The buffer round-trips through world space in F32.** `particle-input-local-block.js` integrates in local space (good) but then writes the reprojected *world* position back into `buffer.posX/posY/posZ`, which `particle-buffer.js` allocates as `Float32Array`. Our ship sits ~1e6 units from its star, where the F32 ULP is ~0.0625 m, against plume sprites 0.16–0.8 m across. So even with a per-frame matrix refresh, the plume would quantize to a ~6 cm world lattice — 8–40 % of a sprite, on the nearest thing on screen — which is the same failure mode as the billboard anchors in #576, one layer down. What our port does instead is keep the buffer in the emitter's local space for the whole particle lifetime, where the numbers stay small, and apply the world transform once on the way to the billboard.

I mention the second one because a matrix-refresh API on its own would look like it solved this and wouldn't, at least not at this scale.

Two more things, offered as information rather than argument, since they may be useful independent of what you decide on the export:

- Lite's `ParticleSystem` really is just struct-of-arrays state plus eight nullable creation slots and an `updateSteps: (i:number)=>void` array, per its own doc comment — no emit shape, no motion integration, no colour ramp. All of that lives in NPE's blocks. So exporting `createParticleSystem` hands out a raw buffer the caller must drive entirely itself, not a rival authoring surface.
- We followed NPE's own field conventions rather than inventing ours, specifically so a future move to a real graph is behaviour-preserving: emit power is folded into `dirX/Y/Z` (a velocity, not a unit vector) and `colorStepR..A` hold `(colorDead - color) / lifeTime`, both per `create-particle-block.js`.

Meanwhile we're not blocked — the port already runs on a local-frame emitter of our own, and a one-line re-export as a fork patch is cheap and stable if the answer stays no. I'd rather have those two gaps on record as real NPE limitations than win the export. Happy to write them up as their own issue so they don't ride on this PR, if that's useful.

VicenteCartas:
> Fair — and §1 is more explicit than I gave it credit for: "it does not expose a Babylon-style particle constructor" is a design requirement, not a description of the current state. I opened this from the barrel's shape rather than the doc's intent, and the PR reads differently once you've read the latter. Treat the export as withdrawn unless you decide otherwise; it isn't the useful part of this.
> 
> The useful part is your question. What's missing turns out to be more specific than we thought. NPE has the capability — two implementation details keep us off it.
> 
> `ParticleInputBlock` with `contextualValue: 0x18` (`LocalPositionUpdated`, routed to `particle-input-local-block.js`) integrates a particle's position in the emitter's local frame on _every_ simulation step, not just at birth — wired through `UpdatePositionBlock` under `SystemBlock.isLocal = true`, that's exactly `isLocal` semantics: local velocity accumulated, reprojected through `emitterWorldMatrix` each call.
> 
> I'd rather give you both blockers than have you fix one and find it didn't help.
> 
> **The emitter frame can't move.** `emitterWorldMatrix` is consumed once, as a `BuildNodeParticleOptions`/`ParseNodeParticleOptions` field, when the graph is built — and the public `Mat4` is `{ readonly length: 16; readonly [index: number]: number }`, so there's no supported way to refresh it for a later frame. Fine for a torch or a campfire; not enough for one riding a moving body. Our hull travels up to 3.2 km/s against a particle lifetime of at most 0.13 s, so an emitter frozen at build time is a few hundred metres from the ship within one particle's life and the plume detaches.
> 
> **The buffer round-trips through world space in F32.** `particle-input-local-block.js` integrates in local space (good) but then writes the reprojected _world_ position back into `buffer.posX/posY/posZ`, which `particle-buffer.js` allocates as `Float32Array`. Our ship sits ~1e6 units from its star, where the F32 ULP is ~0.0625 m, against plume sprites 0.16–0.8 m across. So even with a per-frame matrix refresh, the plume would quantize to a ~6 cm world lattice — 8–40 % of a sprite, on the nearest thing on screen — which is the same failure mode as the billboard anchors in #576, one layer down. What our port does instead is keep the buffer in the emitter's local space for the whole particle lifetime, where the numbers stay small, and apply the world transform once on the way to the billboard.
> 
> I mention the second one because a matrix-refresh API on its own would look like it solved this and wouldn't, at least not at this scale.
> 
> Two more things, offered as information rather than argument, since they may be useful independent of what you decide on the export:
> 
> * Lite's `ParticleSystem` really is just struct-of-arrays state plus eight nullable creation slots and an `updateSteps: (i:number)=>void` array, per its own doc comment — no emit shape, no motion integration, no colour ramp. All of that lives in NPE's blocks. So exporting `createParticleSystem` hands out a raw buffer the caller must drive entirely itself, not a rival authoring surface.
> * We followed NPE's own field conventions rather than inventing ours, specifically so a future move to a real graph is behaviour-preserving: emit power is folded into `dirX/Y/Z` (a velocity, not a unit vector) and `colorStepR..A` hold `(colorDead - color) / lifeTime`, both per `create-particle-block.js`.
> 
> Meanwhile we're not blocked — the port already runs on a local-frame emitter of our own, and a one-line re-export as a fork patch is cheap and stable if the answer stays no. I'd rather have those two gaps on record as real NPE limitations than win the export. Happy to write them up as their own issue so they don't ride on this PR, if that's useful.

Thank you for the detailed explanation! I have captured these items on my backlog for NPE (there's a lot of other stuff missing too), and I will be working through (I'll try to have the emitters part soon, large world may take me a little longer).

Are you ok closing this PR then?

PR #576, opened 2026-08-17
fix(billboard): keep world anchors in F64 so the floating origin can rescue them

VoxelBoy:
## The rule this breaks

`tests/lite/unit/floating-origin-upload.test.ts` states it for the mesh path: at world scale the F32 ULP is coarse (~0.0625 at 1e6), so the eye-relative subtraction has to happen in F64 **before** the F32 store. It asserts that `packMat4IntoF32WithOffset` recovers a 1.5 mm delta from a mesh at 1e6, and that the precision-only packer does not.

Billboard anchors do it in the opposite order. `addBillboardSpriteIndex` stores the caller's position straight into `_instanceData`, a `Float32Array`, and both floating-origin paths in `billboard-pipeline.ts` subtract the camera offset from that already-quantized value:

```js
const sourceData = system._instanceData;   // Float32Array
dest[base]     = sourceData[base]!     - foX;
dest[base + 1] = sourceData[base + 1]! - foY;
dest[base + 2] = sourceData[base + 2]! - foZ;
```

The subtraction is exact, but there is nothing left to subtract from — the sub-ULP detail went one step earlier. A billboard therefore cannot resolve motion finer than the F32 ULP at its own **world** magnitude, however close to the camera it is, which is the thing floating origin exists to prevent. The comment already above that code notes anchors "live at world scale (~5e6)"; at 5e6 the ULP is 0.5 m.

## What it looks like

A sprite snapping to a world-space lattice while the camera moves smoothly past it. The spacing scales with the scene's distance from the world origin, not with the sprite's distance from the eye:

| World magnitude | Lattice | Against a 0.16–0.8 m particle sprite |
| --- | --- | --- |
| 1e6 | 0.0625 m | ~8–40 % of the sprite |
| 5e6 | 0.5 m | comparable to the sprite |

Found porting a ship exhaust plume in a space game: the emitter is the nearest thing to the camera in the scene, a little under 20 m away, while the ship is ~1e6 m from its star — so the plume shimmered against a 6 cm grid despite being the closest object on screen.

## The fix

Adds `_anchor`, an F64 shadow of the anchor positions (3 per sprite), and subtracts from that instead. Maintained exactly as `_savedSize` already is: allocated in `createBillboardSystem`, copied in `growCapacity`, swap-removed in `removeBillboardSpriteIndex`, zeroed in `clearBillboardSprites`.

- `_instanceData` still carries the F32 position. Picking (`billboard-pick-pipeline.ts`) and the bounding centre (`refreshBillboardWorldCenter`) both read it and both work at world scale, where F32 is proportionate.
- `writeInstance` carries a missing `position` forward from `_anchor` rather than from the previous `_instanceData` row, so an update touching only `color` cannot round the anchor down to F32 as a side effect.
- The **non**-floating-origin *upload* path is untouched and keeps its partial dirty-range uploads straight from `_instanceData`.

## Cost, stated plainly

The shadow is allocated for every billboard system, including those that will never see a floating origin. I could not find a way to gate it: a sprite system is created standalone and registered into a scene later, so `createBillboardSystem` has no engine to ask, and the anchor has to be F64-accurate by the time the first sprite is written — earlier than any FO signal is available. If you would rather have the allocation gated, it needs an engine or scene handle at construction, which is an API change I did not want to make on your behalf.

So the honest accounting is:

- **CPU memory:** +24 B per sprite against the existing 64 B instance record, paid by all billboard users. A 10k-sprite system grows from 640 KB to 880 KB.
- **Per-frame work:** none added. The FO path already copied every instance into scratch each frame, because every anchor depends on the live camera offset.
- **Bundle:** +140 to +237 B across the 18 lab scenes that use billboards. All stay within their ceilings; the tightest afterwards is scene118 with 288 B of headroom remaining.

## Test

`tests/lite/unit/billboard-floating-origin-precision.test.ts` puts the two paths side by side at the same magnitude with the same delta, so the gap is a measured number rather than an argument. Before the change, two anchors 1.5 mm apart at 1e6 upload as the **identical** value; the mesh-path case passes throughout, which is what makes it a comparison and not an assertion about F32.

## Checks

- `vitest --project unit` — 1576 passed / 2 skipped, 206 files
- Bundle size — measured on all 18 billboard scenes, no ceiling exceeded
- Per `AGENTS.md`, the all-scene parity suite was not run locally

Second commit documents the shadow in `32-sprites.md`, which described `position` as the world anchor without recording that FO needs an unquantized source for it.

---

Found while porting a WebGPU procedural space game onto Lite, alongside #573 and #574. Each stands alone as a change — the code touches nothing the others touch — but they are **not** file-independent: all three regenerate files under `lab/public/bundle/manifest/` (#573 ∩ #576 = 124, #573 ∩ #574 = 14, #574 ∩ #576 = 7), so whichever lands last needs its manifests regenerated against the new master. See the comment below; this corrects an earlier claim here that the three shared no files.


deltakosh:
Really clean fix. Thanks again:)

Just one ask: Test the carry-forward. The writeInstance change to source a missing position from _anchor instead of prev can be a problem and nothing tests it. A revert to prev would silently reintroduce the snapping on any position-omitting update (say a per-frame alpha tween on a far sprite) and every current test would still pass. Worth one case: add a far sprite, updateBillboardSpriteIndex with only {color}, upload under FO, assert the sub-ULP offset survives.

deltakosh (changes requested):
Test the carry-forward

VoxelBoy:
Done in 6fed793f — one case in `billboard-floating-origin-precision.test.ts`: two sprites at 1e6 and 1e6 + 1.5 mm, a colour-only `updateBillboardSpriteIndex` on the far one, upload under FO, assert the offset survives and the two stay distinct.

I checked it actually catches the regression you described rather than trusting that it would. Reverting `writeInstance`'s fallback from `system._anchor[...]` back to `prev![...]` fails only the new case; the three existing ones stay green, which is the failure mode you predicted:

```
× a position-omitting update carries the anchor forward from `_anchor`, not from the quantized `_instanceData` row
AssertionError: expected +0 to be 0.001500000013038516
```

One thing that came out of writing it, and it makes your concern sharper than I'd stated it in the PR body: at this magnitude the delta doesn't merely lose precision, it vanishes. `Math.fround(1e6 + 0.0015)` is bit-identical to `Math.fround(1e6)`, so `prev` already reads back as plain `FAR`. A single colour-only update on that path wouldn't degrade the anchor, it would overwrite the F64 value with the rounded one and erase the offset permanently, from the first tween frame onward. That's a hard silent failure rather than a gradual one — which is exactly why it needed pinning and why nothing else in the file would have noticed.

I didn't add a second case. A colour-then-position sequence exercises nothing new, since a real position update takes the `props.position` branch and never reads the anchor, and the `growCapacity` interaction is orthogonal and already covered structurally. Happy to add either if you'd rather have them explicit.

`tsc --noEmit` clean on both projects, 45/45 on the billboard unit tests, test-only diff so no bundle movement.

VoxelBoy:
Pushed `de96ca08`, which adds 124 manifest files to what was a four-file PR. That needs explaining, and it's a correction to this PR rather than new work.

**What happened.** #573's Bundle Size job was failing, and the cause turned out to have nothing to do with its review: that PR changed per-scene bundle output without regenerating `lab/public/bundle/manifest/`, and `validate:bundle-manifest` fails a scene once its committed `rawBytes` drifts past a 10 B tolerance. Having found it there, I checked whether the same defect was latent in the other three. It was, in exactly one — this one.

This PR's description has always said the fix adds "+140 to +237 B across the 18 lab scenes that use billboards". That measurement was correct — a full `build:bundle-scenes` confirms exactly 18 scenes in that range, from +140 B (scene98) to +237 B (scene59), net +3056 B. I simply never committed the manifests recording it. 11 further scenes move by 5–6 B either way, inside the tolerance, regenerated for consistency. No ceiling is exceeded; the tightest afterwards is scene118 with 288 B remaining, exactly as the description states. 113/114/115 keep their CI-authored values.

So the Bundle Size job would have failed here the first time it ran. It has not run yet — GitGuardian is still the only check that has ever reported on this PR, including after your `/azp run`, which may itself be worth a glance.

**Correction: the three PRs are not independent.** This description says "no shared files, and each stands alone." That was true when I wrote it and is now false, because all three touch `lab/public/bundle/manifest/`:

| | shared manifest files |
| --- | --- |
| #573 ∩ #576 | 124 |
| #573 ∩ #574 | 14 |
| #574 ∩ #576 | 7 |

#573 is the widest — its allocator change is a small reduction on almost every scene, so it regenerates 222 manifests and subsumes all of this PR's. Whichever of the three lands last will need its manifests regenerated against the new master; the code changes still don't touch each other. The note I left on #573 and #574 mentioned only that pair, since this PR had no manifests at the time — this table supersedes it. Happy to rebase in whatever order suits you.

deltakosh (approved):

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/576/)

<sub>Build 20260818.11 - merge @ 7a43fabfcfa0c046079cdf30f4ba619de3e315d2</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 56 — Axis-Locked Billboards<br/>`scene56` | 59 KB | 58 KB | **+1 KB** |
| Scene 280 — NPE Particles: Flow Map<br/>`scene280` | 42 KB | 41 KB | **+1 KB** |
| Scene 281 — NPE Particles: Noise Texture<br/>`scene281` | 42 KB | 41 KB | **+1 KB** |
| Scene 283 — NPE Particles: Multiply Blend<br/>`scene283` | 42 KB | 41 KB | **+1 KB** |
| Scene 284 — NPE Particles: MultiplyAdd Blend<br/>`scene284` | 42 KB | 41 KB | **+1 KB** |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/576/)

<sub>Build 20260819.18 - merge @ 7a671543d6644a4a7ab1ff9342f10921ccfb6e7d</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 56 — Axis-Locked Billboards<br/>`scene56` | 59 KB | 58 KB | **+1 KB** |
| Scene 98 — Billboard Additive Blend<br/>`scene98` | 59 KB | 58 KB | **+1 KB** |
| Scene 118 — Billboard Sprite Picking<br/>`scene118` | 76 KB | 75 KB | **+1 KB** |
| Scene 280 — NPE Particles: Flow Map<br/>`scene280` | 42 KB | 41 KB | **+1 KB** |
| Scene 283 — NPE Particles: Multiply Blend<br/>`scene283` | 42 KB | 41 KB | **+1 KB** |

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/576/)

<sub>Build 20260819.19 - merge @ 15675bcc0f1887ba93ff6f30c9e25d34475eaa3f</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 56 — Axis-Locked Billboards<br/>`scene56` | 59 KB | 58 KB | **+1 KB** |
| Scene 98 — Billboard Additive Blend<br/>`scene98` | 59 KB | 58 KB | **+1 KB** |
| Scene 118 — Billboard Sprite Picking<br/>`scene118` | 76 KB | 75 KB | **+1 KB** |
| Scene 280 — NPE Particles: Flow Map<br/>`scene280` | 42 KB | 41 KB | **+1 KB** |
| Scene 283 — NPE Particles: Multiply Blend<br/>`scene283` | 42 KB | 41 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260819.19/lab/index.html)

<sub>Build 20260819.19 - merge @ 15675bcc0f1887ba93ff6f30c9e25d34475eaa3f</sub>

PR #583, opened 2026-08-20
feat(compute,mesh): GPU-generated geometry and a user-facing compute seam

VoxelBoy:
Adds a supported path for geometry produced on the GPU and consumed in place, with no readback and no de-interleaving copy, plus the user-facing compute seam that produces it. Discussed in [this forum thread](https://forum.babylonjs.com/t/lite-no-supported-path-for-gpu-generated-geometry-proposal-prototype/63959); shapes follow @deltakosh's answers there.

Split into two independent commits as offered, so the storage-buffer half can land without the compute seam.

## 1. GPU-resident geometry

- `createStorageBuffer(engine, bytes, { writable, vertex, index })` — read_write storage bindings plus VERTEX/INDEX usage. A writable allocation keeps no CPU shadow, since there is nothing meaningful to mirror when the GPU owns the contents and shadowing a large slab doubles its memory, so device-loss rebuild reallocates it empty for the owner to refill.
- `createMeshFromStorageBuffer` — mesh geometry sourced from a `StorageBuffer`, with `_baseVertex` addressing a slot inside a shared slab through the draw call rather than a non-zero `setVertexBuffer` bind offset (which corrupts vertex fetch on some AMD/Dawn paths).
- Borrowed buffers — a storage-backed mesh points every vertex-side field at one allocation shared with every other mesh holding a slot, so disposing one slot no longer destroys the slab the survivors still draw from.

Vertex layout is split across the two things that own it:

- The **mesh** owns the packing (byte stride, per-attribute offset). That is `MeshGPU._vbLayout`, the record the glTF interleave path already produces, reused rather than widened.
- The **material** owns the format, via `setShaderAttributeFormats`. It decides the WGSL type of `input.<attribute>` (e.g. a `float32x4` position packing data in `.w`), which is the shader's own signature and cannot come from whichever mesh is drawn.

That split is what lets one material draw both a tight CPU mesh and a GPU-produced slab. Meshes with different packing cannot share a pipeline, so the opaque renderable groups its packets by `_vbKey`.

## 2. The compute seam

```
createComputeShader / disposeComputeShader
prepareComputeShader        // createComputePipelineAsync, off the frame that dispatches
setComputeUniform / setComputeStorageBuffer
dispatchCompute
```

Shape mirrors `createShaderMaterial`. No raw WebGPU handle crosses the public API — storage binds as an opaque `StorageBuffer` — so pillar 4d holds.

One dispatch path, not two: `compute-pass` records dispatches into a pass and dedups consecutive `setPipeline` calls, and **both** the thin-instance culling batch and this seam go through it, replacing the copy of that loop each previously had. `dispatchCompute` opens its own encoder and submits, as the loaders, the IBL/BRDF preprocessors and the GPU picker already do.

Batching is deliberately absent from this first surface. `queue.writeBuffer` is ordered against submission rather than recording, so the moment dispatches share a command buffer they all observe the last uniform value written and a set of differently-parameterised dispatches silently collapses into N copies of the final one. Re-setting between dispatches is correct here precisely because each dispatch is its own submission. Happy to propose the batched form separately, with that ordering problem solved rather than documented.

## Bundle sizes

Nothing moves. Same-machine A/B against `upstream/master`, both sides built from source, across 18 scenes spanning PBR, Standard, ShaderMaterial, sprites, billboards, thin-instance, VAT and glTF: **15 of 18 byte-identical, max movement 5 bytes**, and every ShaderMaterial scene exactly 0.

Everything that could cost bytes — layout resolution, packet grouping, the slot-aware draw, the borrow-aware disposer, attribute formats — sits behind a module-local hook whose only setter is called by `mesh-from-storage`, the same idiom `_stencilResolver` uses. Without that factory in the bundle the setters tree-shake, the bundler proves the hooks null, and each branch folds back to the expression that was there before.

Formats are a `setShaderAttributeFormats` call rather than a `ShaderMaterialOptions` field for the same reason `setShaderTexture` is a function: carrying the field on the material cost 36 bytes in every ShaderMaterial scene whether or not it declared a format.

## Scene 285

Four terrain chunks generated by one compute dispatch into a single writable, vertex-capable storage buffer, each drawn from its own `baseVertex` slot in that slab, sharing one index allocation, with the colour ramp carried in `position.w`. No readback and no copy on either side.

The Babylon.js reference uses core's `ComputeShader` + `StorageBuffer` created with `BUFFER_CREATIONFLAG_VERTEX` rather than generating the mesh on the CPU, so the parity claim is about compute output on both engines and not just the final image.

**Full-image MAD is 0.000.** Ceiling set tight at 47.48 KB against a measured 47.476 KB.

No draw-call assertion, deliberately: the engines account for this scene differently and both are right. Babylon.js reports 4, one per chunk; Lite reports 1, because four meshes sharing one material and one vertex packing collapse into a single merged binding.

## Checks

- `pnpm lint` clean (eslint + all 6 tsc projects)
- 1676 unit tests pass
- scene 285 parity passes at MAD 0.000
- rebased onto `46cc8631`

A second PR follows for ShaderMaterial support of interleaved data, as requested in the thread.


bjsplat:
## API Changes

<details>

```diff
diff --git a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
index 64d2b908..3b2f2b97 100644
--- a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md
+++ b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
@@ -1124,6 +1124,47 @@ export function computeMaxExtents(meshes: readonly Mesh[], animationGroup?: Anim
 // @public
 export function computePath(plugin: NavigationPlugin, start: Vec3, end: Vec3): Vec3[];
 
+// @public
+export interface ComputeShader {
+    // (undocumented)
+    readonly [computeShaderBrand]: true;
+    // (undocumented)
+    readonly name: string;
+}
+
+// @public
+export interface ComputeShaderOptions {
+    readonly computeSource: string;
+    // (undocumented)
+    readonly defines?: ShaderDefineMap;
+    readonly entryPoint?: string;
+    // (undocumented)
+    readonly name?: string;
+    // (undocumented)
+    readonly storageBuffers?: readonly ComputeStorageBufferDecl[];
+    // (undocumented)
+    readonly uniforms?: readonly ComputeUniformDecl[];
+}
+
+// @public
+export interface ComputeStorageBufferDecl {
+    // (undocumented)
+    readonly name: string;
+    // (undocumented)
+    readonly type: string;
+    readonly writable?: boolean;
+}
+
+// @public
+export interface ComputeUniformDecl {
+    // (undocumented)
+    readonly defaultValue?: ShaderUniformValue;
+    // (undocumented)
+    readonly name: string;
+    // (undocumented)
+    readonly type: ShaderUniformType;
+}
+
 // @public
 export function computeYawPitchFromLookAt(
     lookAt: Vec3,
@@ -1289,6 +1330,9 @@ export function createClusteredLightContainer(options?: ClusteredLightContainerO
 // @public
 export function createClusteredPointLight(container: ClusteredLightContainer, options: ClusteredPointLightOptions): ClusteredPointLight;
 
+// @public
+export function createComputeShader(engine: EngineContext, options: ComputeShaderOptions): ComputeShader;
+
 // @public
 export function createCopyToTextureTask(config: CopyToTextureTaskConfig, engine: EngineContext, scene: SceneContext): CopyToTextureTask;
 
@@ -1479,6 +1523,9 @@ export function createMeshFromData(
     colors?: Float32Array
 ): Mesh;
 
+// @public
+export function createMeshFromStorageBuffer(engine: EngineContext, name: string, options: MeshFromStorageOptions): Mesh;
+
 // @public
 export function createMicrophoneSoundSourceAsync(engine: AudioEngine, options?: SoundSourceOptions): Promise<AudioInputSource>;
 
@@ -1709,7 +1756,7 @@ export function createStandardMaterial(): StandardMaterialProps;
 export function createStandardNoColorMaterialView(source: StandardMaterialProps): MaterialView;
 
 // @public
-export function createStorageBuffer(engine: EngineContext, data: ArrayBufferView, label?: string): StorageBuffer;
+export function createStorageBuffer(engine: EngineContext, source: ArrayBufferView | number, labelOrOptions?: string | StorageBufferOptions): StorageBuffer;
 
 // @public
 export function createStreamingSoundAsync(engine: AudioEngine, source: StreamingSoundSource, options?: StreamingSoundOptions): Promise<StreamingSound>;
@@ -2126,6 +2173,9 @@ export interface DiscOptions {
     tessellation?: number;
 }
 
+// @public
+export function dispatchCompute(engine: EngineContext, shader: ComputeShader, x: number, y?: number, z?: number): void;
+
 // @public
 export function disposeAudioBus(bus: AudioBus): void;
 
@@ -2150,6 +2200,9 @@ export function disposeBoundingBoxGizmo(gizmo: BoundingBoxGizmo, layer: UtilityL
 // @public
 export function disposeCameraGizmo(gizmo: CameraGizmo, layer: UtilityLayer): void;
 
+// @public
+export function disposeComputeShader(shader: ComputeShader): void;
+
 // @public
 export function disposeCsg2(solid: Csg2Solid): void;
 
@@ -3661,6 +3714,20 @@ export interface MeshExtent {
     minimum: [number, number, number];
 }
 
+// @public
+export interface MeshFromStorageOptions {
+    readonly arrayStride: number;
+    readonly attributeOffsets?: Partial<Record<ShaderAttributeName, number>>;
+    readonly baseVertex?: number;
+    readonly boundMax?: readonly [number, number, number];
+    readonly boundMin?: readonly [number, number, number];
+    readonly indexCount?: number;
+    readonly indexFormat?: GPUIndexFormat;
+    readonly indices: Uint16Array | Uint32Array | StorageBuffer;
+    readonly storage: StorageBuffer;
+    readonly vertexCount: number;
+}
+
 // @public (undocumented)
 export interface MeshGeometryCapacityResult {
     // (undocumented)
@@ -4857,6 +4924,9 @@ export function preloadStreamingInstanceAsync(sound: StreamingSound): Promise<vo
 // @public
 export function preloadStreamingInstancesAsync(sound: StreamingSound, count: number): Promise<void>;
 
+// @public
+export function prepareComputeShader(shader: ComputeShader): Promise<void>;
+
 // @public
 export interface PreparedVatBakeResult {
     // (undocumented)
@@ -5624,6 +5694,12 @@ export function setCameraLimits(camera: ArcRotateCamera, limits: ArcRotateCamera
 // @public
 export function setClipPlane(scene: SceneContext, plane: ClipPlane): void;
 
+// @public
+export function setComputeStorageBuffer(shader: ComputeShader, name: string, buffer: StorageBuffer): void;
+
+// @public
+export function setComputeUniform(shader: ComputeShader, name: string, value: ShaderUniformValue): void;
+
 // @public
 export function setDracoBaseUrl(url: string): void;
 
@@ -5792,6 +5868,9 @@ export function setScaleGizmoLocalCoordinates(gizmo: ScaleGizmo, useLocal: boole
 // @public
 export function setSceneImageProcessing(scene: SceneContext, update: ImageProcessingUpdate): Promise<void>;
 
+// @public
+export function setShaderAttributeFormats(material: ShaderMaterial, formats: ShaderAttributeFormats): void;
+
 // @public
 export function setShaderFloat(material: ShaderMaterial, name: string, value: number): void;
 
@@ -5935,6 +6014,9 @@ export function setVatInstanceStorage(engine: EngineContext, mesh: Mesh, buffer:
 // @public
 export function setVatTime(engine: EngineContext, mesh: Mesh, seconds: number): void;
 
+// @public
+export type ShaderAttributeFormats = Partial<Record<ShaderAttributeName, GPUVertexFormat>>;
+
 // @public
 export type ShaderAttributeName = "position" | "normal" | "uv" | "uv2" | "tangent" | "color" | "joints" | "weights" | "joints1" | "weights1";
 
@@ -6677,6 +6759,15 @@ export interface StorageBuffer {
     readonly byteLength: number;
 }
 
+// @public
+export interface StorageBufferOptions {
+    readonly index?: boolean;
+    // (undocumented)
+    readonly label?: string;
+    readonly vertex?: boolean;
+    readonly writable?: boolean;
+}
+
 // @public
 export interface StreamingSound {
     readonly instanceCount: number;
```

</details>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/583/)

<sub>Build 20260820.8 - merge @ 62c3033a33c7bac5b8464c2894b03887d735b832</sub>

deltakosh (changes requested):
Changes requested:

- Preserve INDEX and COPY_SRC usage flags during storage-buffer device-loss rebuilds.
- Apply baseVertex consistently to PBR, Standard, and picking indexed draws.
- Map packed uint/sint vertex formats to u32/i32 WGSL inputs.
- Recreate compute pipelines, layouts, UBOs, and bind groups after device loss.
- Refresh storage-backed mesh GPU handles after storage-buffer rebuilds.

RaananW:
@VoxelBoy - just making sure you saw the change requests :)

RaananW (changes requested):
**Verdict: Request changes**

**CI: Failing** — lint/typecheck, unit, parity, API, compat, performance, and release checks passed, but PR Validation Bundle Size failed. The failure is from the obsolete committed-manifest validation; the actual ceiling check was skipped, so this head has no valid bundle-size signal and needs a rebase/re-run.

The overall design is strong: material-owned formats and mesh-owned packing are the right split, borrowed GPU-buffer ownership is explicit, and the compute-pass extraction removes duplication. I found these blocking correctness issues:

1. `shader-vb.ts` stores declared formats per material, but `shader-pipeline-cache.ts:121` does not include them in `bindingsKey`. Multi-ShaderMaterial groups automatically enable the cross-material cache, so materials with identical declarations but different formats can share one `ShaderPipelineBindings` and vertex layout. That can produce pipeline validation errors or silently garbled vertex fetch. Include the declared formats in both the bindings and variant cache keys.

2. `storage-buffer.ts:185` rebuilds buffers with only `STORAGE | VERTEX`, dropping `INDEX` and writable buffers' `COPY_SRC`. After device loss, index-backed allocations fail validation and writable allocations no longer match their creation usage. Reuse the complete creation usage expression.

3. `_baseVertex` is honored only in the ShaderMaterial draw path (`shader-vb.ts:154`). PBR, Standard, and GPU picking all draw without it even though they accept the same mesh and PBR already honors `_vbLayout`. A slab mesh with `baseVertex > 0` therefore draws/picks slot 0 outside ShaderMaterial. Apply `_baseVertex` consistently or reject unsupported material/picking paths.

4. `wgslTypeForFormat` maps packed integer formats such as `uint8x4` and `sint16x2` to `vec4<f32>`, but WebGPU requires integer shader inputs for non-normalized integer formats. Handle the `uint*`/`sint*` families explicitly.

5. `MeshFromStorageOptions.attributeOffsets` advertises tangent/uv2/color, and `_vbLayout` records them, but the mesh only points position/normal/uv at the shared slab. Those other attributes fall back to tiny zero buffers while retaining the slab stride, causing wrong reads or buffer-size validation failures. Wire all advertised streams to the slab or narrow and enforce the supported set.

6. `ensureBindGroup` caches handles without detecting that a bound `StorageBuffer` was disposed or rebuilt, bypassing `_getStorageBufferHandle` on later dispatches. `ComputeShader` also has no engine device-loss lifecycle, leaving its pipeline/layout/UBO tied to the old device. Track buffer generations and rebuild/invalidate compute GPU state with the engine lifecycle.

Process cleanup is also required on rebase: remove `lab/public/bundle/manifest/scene285.json` and the associated ignore exceptions, since master no longer tracks generated bundle manifests. The current `maxRawKB` leaves about four bytes of headroom; any wider ceiling needs explicit approval rather than an implicit threshold change. Please also replace the module-scope `new WeakMap()` with the lazy-init form required by GUIDANCE.md and add focused tests for the cache-key, rebuild-usage, packed-integer, and baseVertex paths.

Popov72:
Thanks for putting this together.

However, I think the compute half needs to be reworked before it becomes public API. The current implementation conflates three different things:

1. the compiled compute program;
2. one mutable set of bound resources and uniforms;
3. the scheduling/submission of one invocation.

That ownership model is the root of most of the limitations below.

## `ComputeShader` should represent the program, not one mutable invocation

It makes sense for `ComputeShader` to own program-level state: WGSL, entry point, bind-group layouts, pipeline variants, asynchronous compilation state, and device-rebuild state.

It should not own the only UBO, the only mutable resource map, and the only bind group. With the current shape, changing resources destroys the previous binding state, A/B/A resource combinations cannot be reused, multiple uniform buffers cannot be bound, and every differently parameterized dispatch is forced through the same mutable UBO.

I think the responsibilities should instead be separated approximately as follows:

- `ComputeShader`: immutable program/pipeline/layout state;
- `ComputeBindingSet`: a reusable, prevalidated combination of buffers, textures and samplers;
- `ComputeDispatch`: shader + binding set + direct/indirect workgroup arguments + per-dispatch uniform offset;
- `ComputeTask` / `ComputePass`: the ordered reusable dispatch list and its recording into the frame encoder.

An opaque `UniformBuffer` resource, following the same standalone create/update/dispose style as `StorageBuffer`, would also be useful. A compute task could additionally provide a uniform-ring convenience path for small per-dispatch values.

## A compute task is a scheduling point, not one task per dispatch

`dispatchCompute` currently creates one command encoder, one compute pass, one command buffer, and one `queue.submit` per dispatch. Therefore 200 dispatches in a frame produce 200 submissions (the [Ocean Demo](https://playground.babylonjs.com/?webgpu#YX6IB8#758) has around 200 dispatches per frame). The pipeline-deduplication logic in `recordComputeDispatches` does not help the public path because it is always called with a one-element list.

The frame graph is already designed as the ordered scheduler for GPU work, and its `Pass` abstraction explicitly anticipates compute passes. A single `ComputeTask` should be able to contain 200 dispatches and record them into one compute pass on the frame encoder. Additional compute tasks/passes should only be needed at real ordering boundaries, for example:

```text
compute simulation (120 dispatches)
depth render
depth-dependent compute (75 dispatches)
main render
post-render compute (5 dispatches)
```

This also makes compute ordering relative to render/copy work explicit. The immediate-submission API cannot correctly express "run after this render task but before that render task"; a separately submitted command buffer jumps ahead of render work already encoded but not yet submitted.

`dispatchCompute` can remain as a convenience for genuinely one-off initialization work, but it should not be the only public execution model.

## The uniform-write problem does not require one submission per dispatch

The observation about `queue.writeBuffer` is correct for the current single mutable UBO: if that buffer is rewritten several times before one command buffer is submitted, all dispatches observe the final contents.

However, this is not a WebGPU limitation and should not be used as the architectural reason to omit batching. The standard solution is immutable per-dispatch storage, typically an aligned uniform ring plus a dynamic offset:

```ts
pass.setBindGroup(0, bindingSet.bindGroup, [dispatch.uniformOffset]);
pass.dispatchWorkgroups(dispatch.x, dispatch.y, dispatch.z);
```

Separate dispatches in a compute pass are separate WebGPU usage scopes, so read-after-write and write-after-write ordering between consecutive dispatches is already defined. The problem to solve is uniform aliasing, not dispatch synchronization.

For that reason, I would also remove the historical "we hit exactly that while prototyping" wording and document the objective constraint and chosen mechanism instead.

## The declared resource surface is too narrow for a general `ComputeShader`

The current API supports only one generated group containing one generated UBO and whole-buffer storage bindings. At minimum, a general compute surface should account for:

- sampled textures and sampler declarations;
- storage textures, including format and access;
- more than one uniform-buffer binding;
- storage-buffer subranges (`offset` / `size`);
- dynamic uniform/storage offsets;
- direct and indirect dispatch;
- pipeline override constants;
- reusable binding sets and, eventually, multiple bind groups.

Indirect dispatch is particularly relevant to the stated GPU-owned/no-readback goal: compute commonly produces the next dispatch or draw count on the GPU. `StorageBufferOptions` currently cannot even request `GPUBufferUsage.INDIRECT`.

The declaration model should also avoid introducing another independent `type: string` surface where an existing shared declaration can be reused. Arbitrary WGSL storage element types do need an open string, but the common declaration shape should still be shared rather than duplicated.

## The steady-state path should be prepared once

The intended frame path should not perform repeated declaration searches or allocate temporary containers.

### UBO / uniform updates

The following points specifically concern the generated UBO and `setComputeUniform` path.

More fundamentally, a `UniformBuffer` resource should be introduced under `resource/`, following the same opaque interface plus standalone create/update/dispose-function model as `StorageBuffer`. The UBO should be owned independently of `ComputeShader` so binding sets can reference one or more uniform buffers, share them between shaders where layouts permit, and manage their lifetime explicitly. A compute task can still provide a uniform-ring convenience path for per-dispatch values.

- precompute uniform slots and layout metadata by name;
- retain the `DataView` or typed views instead of creating one per setter call;
- provide specialized, allocation-free setters for the standard WGSL types (`i32`, `u32`, `f32`, `f16`, vectors, and matrices), so hot code does not repeat dynamic type/shape checks or generic array conversion for every update; keep the generic validated setter as a convenience/setup API;
- validate the exact uniform element count so a value cannot overwrite the next field;
- do not compare against the previous value in these setters: specialized setters should write and mark the UBO dirty directly, and callers are responsible for avoiding redundant updates when the value has not changed.

### Resource binding and dispatch recording (not UBO-related)

These are separate costs in resource binding, command recording, and the immediate-dispatch API:

- precompute resource-binding slots by name;
- make rebinding the same resource a no-op;
- reuse dispatch records and binding sets;
- avoid the dispatch object and one-element arrays currently allocated by every `dispatchCompute`;
- remove the redundant `engine` parameter from `dispatchCompute`, since the shader already owns its engine;
- allow WebGPU's valid zero-sized no-op dispatch, while rejecting fractional/non-finite counts correctly.

Validation should remain on the public string-based setup/update APIs in both categories; the optimization is to precompute validation metadata and avoid repeating setup work, not to silently accept invalid names or value shapes.

There is also a UBO-specific correctness issue: `setComputeUniform` can currently overwrite adjacent fields. Separately, and unrelated to the UBO path, `prepareComputeShader` can race a dispatch or disposal and create/retain a second pipeline.
## Suggested direction

Conceptually, I would expect usage closer to:

```ts
const shader = createComputeShader(engine, shaderDescriptor);
const bindingsA = createComputeBindingSet(shader, resourcesA);
const bindingsB = createComputeBindingSet(shader, resourcesB);
const task = createComputeTask(engine, "simulation");

addComputeDispatch(task, shader, bindingsA, dispatchA);
addComputeDispatch(task, shader, bindingsB, dispatchB);
addTaskBefore(scene, task, renderTask);
```

The task can then prepare pipelines/binding sets outside the frame loop, upload all per-dispatch uniform data once, and record every dispatch into one pass with only necessary pipeline/bind-group changes.

Because this is a new public API, landing the single-UBO/immediate-submit model now and adding batching later is risky: separating program state, binding state and scheduling afterward would require either breaking the API or retaining the inefficient path indefinitely. I think that separation should happen before the compute seam lands.

VoxelBoy:
Fair point, and the lock-in argument settles it for me. If `createComputeShader`
and `dispatchCompute` go public in this shape, separating program state from
binding state from scheduling later means either breaking the API or keeping the
immediate-submit path around forever. Not worth it.

Two things already fixed here so you don't re-review them. The
`prepareComputeShader` race is gone: it grabs the device before the await and
re-checks device and disposal after, so a loss mid-compile can't land a
dead-device pipeline any more. That turned out to be the same bug @RaananW hit
from the device-loss side. `setComputeUniform` overwriting neighbouring fields is
still there.

What I'd like to do is split this and pull the compute seam out.

The storage half stays: `createStorageBuffer`, `createMeshFromStorageBuffer`,
borrowed buffer ownership, mesh-owned packing against material-owned formats,
and device-loss recovery for all of it. Nothing in your review touches that part,
and @RaananW's items on it are done or nearly. I'd like to finish it and land it.

The compute seam I'd rather withdraw than reshape in place. Being straight about
why: what you're describing is a good deal bigger than what we proposed, and most
of the open questions in it are yours rather than ours. How compute tasks order
against render tasks in the frame graph, whether the uniform ring belongs to the
task or to the engine, those are engine architecture calls. I don't think we
should be settling them by writing an implementation first and having you review
the result. If you want us to build it once the design is settled, say so and
we'll scope it properly. We're not assuming that either way. What we can usefully
offer regardless is being the first real consumer, since Manywhere generates its
terrain through this and gives you a live workload to check a design against.

One thing I need to know before splitting. Scene 285 exercises both halves at
once, since compute writes the slab it then draws, so it follows the compute work
out and the storage half is left without a parity scene. Does it need one? If so
I'll add a scene that fills a vertex-capable `StorageBuffer` from the CPU and
draws a few meshes out of one slab through `baseVertex`. That covers the storage
contract without involving compute at all.

I'll leave the rest of the compute-side cleanup for now. The architecture docs
@deltakosh asked for describe exactly the API under discussion.

Popov72:
You’re right — the scheduling, frame-graph integration, and uniform ownership questions are engine architecture decisions that we should settle first. Please split the PR and remove the compute shader support; we’ll take care of designing and implementing that surface. There’s no need to continue the compute-side cleanup or documentation in this PR. Manywhere would still be very useful as a real-world consumer when we validate the eventual design.

The storage-buffer half remains valuable and can proceed independently.

Regarding scene 285: yes, please keep parity coverage by replacing it with the storage-only scene you described. Initializing a vertex-capable `StorageBuffer` from the CPU and drawing several meshes from a shared slab using non-zero `baseVertex`—and the shared index allocation—would cover the important storage contract without coupling it to compute. This end-to-end scene is useful because the unit tests alone cannot validate the real WebGPU buffer usages, vertex layout, and indexed draw integration.

VoxelBoy:
Split done. The compute seam is out — no `compute/` module, no `createComputeShader`
or `dispatchCompute`, and the `thin-instance-gpu-culling` refactor that came with it
is reverted to master's. The public surface this PR now adds is just
`createStorageBuffer` / `updateStorageBuffer` / `disposeStorageBuffer`,
`createMeshFromStorageBuffer`, and `setShaderAttributeFormats`.

The branch is rebuilt on current master rather than rebased, so the force-push looks
larger than it is. @deltakosh's review commit is still there with his authorship, at
a new SHA — the shader half of it moved down into this PR, since the code it fixes is
introduced here and @RaananW raised those items against this PR. Its glTF loader half
stays on #584.

**Scene 285 is replaced**, as you asked, with the storage-only version: a
vertex-capable `StorageBuffer` seeded from the CPU, four meshes drawing slots out of
one slab through a non-zero `baseVertex`, one shared index allocation behind all of
them, and a `float32x4` position whose `.w` carries a per-vertex value the shader
reads back out. Same geometry as before, so the picture is unchanged. Parity passes
at MAD 0.000 against a freshly captured Babylon.js reference.

**The ceiling question went away with the compute seam.** Without it the scene
measures 43.80 KB against the 47.48 KB it was authored with, so rather than the raise
I was asking approval for, the ceiling comes *down* to 44.05 KB with 256 B of margin.
Nothing left to approve.

No golden is committed for it, deliberately. Goldens have to be captured under
SwiftShader as CI renders; a GPU capture is what produced the MAD 0.195 failure on
#584, and I've dropped that one there too pending a CI capture — details in that
thread.

Remaining from @RaananW's review, all in this half:

- rebuild usage flags now reproduce the exact creation mask rather than recomputing
- `baseVertex` reaches PBR, Standard, both picking paths and the ShaderMaterial
  thin-instance path, which had been missed
- `attributeOffsets` binds every stream it advertises; a stream that isn't bound no
  longer gets a layout entry describing a buffer it doesn't read
- storage-backed meshes are re-pointed after a device-loss rebuild, through a weak
  registry so nothing is pinned for the page's lifetime
- the packed-integer WGSL types, bindings cache key and module-level `WeakMap` come
  from @deltakosh's commit
- focused tests for each, including the baseVertex coverage you asked for

The compute items are dropped as you said: no `setComputeUniform` validation, no
compute architecture docs. Happy to be the first consumer whenever you want the
eventual design exercised — Manywhere drives terrain generation through this, so
there's a real workload to point at it.

RaananW (changes requested):
🤖 This review comment was created by an AI agent.

Verdict: 🔴 Request changes
CI: ⚠️ Missing coverage. Only GitGuardian reports on this head, and the PR conflicts with `master`.
CI safety: ✅ No credential-exposure risk found.

Thank you for splitting out the compute API and addressing the earlier cache, usage, base-vertex, format, and ownership concerns. Two blockers remain:

1. [`refreshSlabMeshes`](https://github.com/BabylonJS/Babylon-Lite/blob/e029fc87f5362cdd9d4f4711975fc740c54e0e5d/packages/babylon-lite/src/mesh/mesh-from-storage.ts#L220-L227) repoints position, normal, and UV after device loss, but leaves slab-backed tangent, UV2, and color fields on the dead GPU buffer. Please repoint every optional slab-backed field and add device-loss coverage.
2. [`scene285-storage-backed-geometry.spec.ts`](https://github.com/BabylonJS/Babylon-Lite/blob/e029fc87f5362cdd9d4f4711975fc740c54e0e5d/tests/lite/parity/scenes/scene285-storage-backed-geometry.spec.ts#L6-L13) has no committed `babylon-ref-golden.png`. It therefore regenerates ground truth during CI. Please commit the one-time approved golden and compare only against it.

Please also rebase and run the full required CI.

RaananW (dismissed):
🤖 This review comment was created by an AI agent.

Verdict: 🔴 Request changes

The current head is unchanged since my September 11 review, so its two blockers and missing required CI still apply. I found one additional device-loss blocker: [`createMeshFromStorageBuffer`](https://github.com/BabylonJS/Babylon-Lite/blob/e029fc87f5362cdd9d4f4711975fc740c54e0e5d/packages/babylon-lite/src/mesh/mesh-from-storage.ts#L171-L172) retains no rebuild source when `indices` is a typed array. `refreshSlabMeshes` only replaces shared `StorageBuffer` indices, and normal mesh recovery skips storage-backed meshes because they have no CPU positions. After recovery, the mesh still draws with an index buffer from the lost device. Please retain and recreate typed-array indices on the replacement device, and add a full-recovery test that confirms a new-device index buffer.

CI currently reports only GitGuardian; build, lint/typecheck, unit, parity, API, bundle-size, compatibility, performance, and release checks are absent. The PR also conflicts with `master`. Please rebase and run the full required CI.

VoxelBoy:
@RaananW all three device-loss and golden blockers are fixed, and the branch is rebased on current master.

**Slab-backed streams after device loss.** You were right, and it went a step further: the test covering the owned-index case was asserting the bug, on the assumption that the mesh's own recovery restores it. It doesn't. `_rebuildMeshes` skips anything without CPU positions, so the storage rebuild observer is the only recovery these meshes get. It now:

- re-points every field holding the dead slab handle, by identity rather than a field list, so tangent/uv2/color can't be missed again
- keeps a typed-array `indices` (by reference, as `createMeshFromData` keeps `_cpuIndices`) and uploads it on the replacement device
- drops a disposed mesh's registry entry, so recovery never hands it a buffer nothing would destroy

The new tests cover each of those, plus a full `runDeviceLostRecovery` followed by `_rebuildMeshes` that checks every handle, index buffer included, came from the replacement device. Three of them fail against the previous code.

**Scene 285 golden.** Committed, and the spec no longer calls `captureGolden`, so a missing golden fails instead of being captured mid-run. That's the same form as the newest parity specs.

I also need to correct my last update, where I said goldens had to be captured under SwiftShader. That's wrong: SwiftShader is only the lite-gl job. WebGPU parity runs on BrowserStack macOS Chrome with real WebGPU. What actually matters is that this image already passed there: the committed golden is pixel-identical to the reference CI compared against on build 20260907.1, where Lite measured MAD 0.0056 against the 0.01 limit.

**Rebase.** It surfaced a regression that wasn't visible before. @deltakosh's bindings-key fix read the declared formats through a value import from `shader-pipeline`, and on current master that pulls ShaderMaterial's pipeline chunks into bundles that never use it. scene267 grew 281 B and went over its ceiling. The key now reads `material._attributeFormats` directly, which is equivalent: with no mesh packing, the layout depends only on `attributes` (already keyed) and those formats. scene267 and scene20 are back to +7 B, which is the `baseVertex` argument. A source check keeps the import from coming back. His commit is unchanged.

**Scene 285's ceiling goes from 44.05 to 44.35 KB.** It measures 44.094 KB. Master moved it 259 B on its own, 3 B past the old ceiling before any change here, and the device-loss fix adds 48 B. The new ceiling keeps the same ~256 B of margin it was authored with.

**CI.** The only pipeline run on this PR was @deltakosh's `/azp run` on August 20, so the full required run needs one from one of you.

bjsplat:
## API Changes

**Potentially breaking changes detected.** Removed or changed public API lines:

- `readonly attributes: readonly ShaderAttributeName[];`
- `readonly blend?: GPUBlendState;`
- `readonly blendMode: "alpha" | "additive";`
- `readonly defines: readonly ShaderDefine[];`
- `readonly depthBias: number;`
- `readonly depthBiasSlopeScale: number;`
- `readonly depthCompare: GPUCompareFunction;`
- `readonly depthOnlyFragment: boolean;`
- `readonly depthWrite: boolean;`
- `readonly fragmentSource: string;`
- `readonly name?: string;`
- `readonly needAlphaBlending: boolean;`
- `readonly needAlphaTesting: boolean;`
- `stencil?: StencilState;`
- `readonly transmissive: boolean;`
- `readonly uniformDecls: readonly ShaderUniformDecl[];`

<details>

```diff
diff --git a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
index a3b73018..5cd6c9c7 100644
--- a/home/vsts/work/1/s/test-results/api-report/target/temp/babylon-lite.api.md
+++ b/home/vsts/work/1/s/test-results/api-report/current/temp/babylon-lite.api.md
@@ -1673,6 +1673,9 @@ export function createMeshFromData(
     colors?: Float32Array
 ): Mesh;
 
+// @public
+export function createMeshFromStorageBuffer(engine: EngineContext, name: string, options: MeshFromStorageOptions): Mesh;
+
 // @public
 export function createMicrophoneSoundSourceAsync(engine: AudioEngine, options?: SoundSourceOptions): Promise<AudioInputSource>;
 
@@ -1924,7 +1927,7 @@ export function createStandardMaterial(): StandardMaterialProps;
 export function createStandardNoColorMaterialView(source: StandardMaterialProps): MaterialView;
 
 // @public
-export function createStorageBuffer(engine: EngineContext, data: ArrayBufferView, label?: string): StorageBuffer;
+export function createStorageBuffer(engine: EngineContext, source: ArrayBufferView | number, labelOrOptions?: string | StorageBufferOptions): StorageBuffer;
 
 // @public
 export function createStreamingSoundAsync(engine: AudioEngine, source: StreamingSoundSource, options?: StreamingSoundOptions): Promise<StreamingSound>;
@@ -4486,6 +4489,20 @@ export interface MeshExtent {
     minimum: [number, number, number];
 }
 
+// @public
+export interface MeshFromStorageOptions {
+    readonly arrayStride: number;
+    readonly attributeOffsets?: Partial<Record<ShaderAttributeName, number>>;
+    readonly baseVertex?: number;
+    readonly boundMax?: readonly [number, number, number];
+    readonly boundMin?: readonly [number, number, number];
+    readonly indexCount?: number;
+    readonly indexFormat?: GPUIndexFormat;
+    readonly indices: Uint16Array | Uint32Array | StorageBuffer;
+    readonly storage: StorageBuffer;
+    readonly vertexCount: number;
+}
+
 // @public (undocumented)
 export interface MeshGeometryCapacityResult {
     // (undocumented)
@@ -6950,6 +6967,9 @@ export function setScaleGizmoLocalCoordinates(gizmo: ScaleGizmo, useLocal: boole
 // @public
 export function setSceneImageProcessing(scene: SceneContext, update: ImageProcessingUpdate): Promise<void>;
 
+// @public
+export function setShaderAttributeFormats(material: ShaderMaterial, formats: ShaderAttributeFormats): void;
+
 // @public
 export function setShaderFloat(material: ShaderMaterial, name: string, value: number): void;
 
@@ -7102,6 +7122,9 @@ export function setVatInstanceStorage(engine: EngineContext, mesh: Mesh, buffer:
 // @public
 export function setVatTime(engine: EngineContext, mesh: Mesh, seconds: number): void;
 
+// @public
+export type ShaderAttributeFormats = Partial<Record<ShaderAttributeName, GPUVertexFormat>>;
+
 // @public
 export type ShaderAttributeName = "position" | "normal" | "uv" | "uv2" | "tangent" | "color" | "joints" | "weights" | "joints1" | "weights1";
 
@@ -7122,41 +7145,25 @@ export type ShaderDefineValue = boolean | number;
 // @public
 export interface ShaderMaterial extends Material {
     // (undocumented)
-    readonly attributes: readonly ShaderAttributeName[];
-    // (undocumented)
+    readonly attributes: readonly ShaderAttributeName[]; // (undocumented)
     readonly backFaceCulling: boolean;
-    readonly blend?: GPUBlendState;
-    // (undocumented)
-    readonly blendMode: "alpha" | "additive";
-    // (undocumented)
-    readonly defines: readonly ShaderDefine[];
-    // (undocumented)
-    readonly depthBias: number;
-    // (undocumented)
-    readonly depthBiasSlopeScale: number;
-    // (undocumented)
-    readonly depthCompare: GPUCompareFunction;
-    // (undocumented)
-    readonly depthOnlyFragment: boolean;
-    // (undocumented)
-    readonly depthWrite: boolean;
-    // (undocumented)
-    readonly fragmentSource: string;
-    // (undocumented)
-    readonly name?: string;
-    // (undocumented)
-    readonly needAlphaBlending: boolean;
-    // (undocumented)
-    readonly needAlphaTesting: boolean;
-    // (undocumented)
+    readonly blend?: GPUBlendState; // (undocumented)
+    readonly blendMode: "alpha" | "additive"; // (undocumented)
+    readonly defines: readonly ShaderDefine[]; // (undocumented)
+    readonly depthBias: number; // (undocumented)
+    readonly depthBiasSlopeScale: number; // (undocumented)
+    readonly depthCompare: GPUCompareFunction; // (undocumented)
+    readonly depthOnlyFragment: boolean; // (undocumented)
+    readonly depthWrite: boolean; // (undocumented)
+    readonly fragmentSource: string; // (undocumented)
+    readonly name?: string; // (undocumented)
+    readonly needAlphaBlending: boolean; // (undocumented)
+    readonly needAlphaTesting: boolean; // (undocumented)
     readonly samplerDecls: readonly ShaderSamplerDecl[];
-    stencil?: StencilState;
-    // (undocumented)
+    stencil?: StencilState; // (undocumented)
     readonly storageBufferDecls: readonly ShaderStorageBufferDecl[];
-    readonly transmissive: boolean;
-    // (undocumented)
-    readonly uniformDecls: readonly ShaderUniformDecl[];
-    // (undocumented)
+    readonly transmissive: boolean; // (undocumented)
+    readonly uniformDecls: readonly ShaderUniformDecl[]; // (undocumented)
     readonly vertexSource: string;
 }
 
@@ -7930,6 +7937,15 @@ export interface StorageBuffer {
     readonly byteLength: number;
 }
 
+// @public
+export interface StorageBufferOptions {
+    readonly index?: boolean;
+    // (undocumented)
+    readonly label?: string;
+    readonly vertex?: boolean;
+    readonly writable?: boolean;
+}
+
 // @public
 export interface StreamingSound {
     readonly instanceCount: number;
```

</details>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/583/)

<sub>Build 20260914.3 - merge @ 4e3ec035f0fbefc6bc5d04afff81f60f41647d53</sub>

bjsplat:
## Bundle Size Changes

No changes at whole-KB resolution — but this PR put a scene close to its ceiling.

### Ceiling headroom

🚨 **3 scenes put over their ceiling by this PR:** `scene163` (+16 B over its 39.10 KB ceiling), `scene159` (+3 B over its 39.38 KB ceiling), `scene162` (+1 B over its 39.40 KB ceiling)

⚠️ **69 scenes this PR added or grew now sit under 1.0 KB of headroom.**

| Scene | Size | Ceiling | Headroom | Δ this PR |
|-------|------|---------|----------|-----------|
| Scene 161 - ShaderMaterial Uniform<br/>`scene161` | 39.91 KB | 39.94 KB | **33 B** | +86 B |
| Scene 160 - ShaderMaterial Texture<br/>`scene160` | 41.21 KB | 41.28 KB | **74 B** | +83 B |
| Scene 213 — GridMaterial<br/>`scene213` | 49.43 KB | 49.52 KB | **94 B** | +92 B |
| Scene 16 — Thin Instances<br/>`scene16` | 48.18 KB | 48.30 KB | **122 B** | +10 B |
| Scene 19 — PBR Clearcoat<br/>`scene19` | 69.17 KB | 69.30 KB | **133 B** | +7 B |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 44.66 KB | 44.80 KB | **143 B** | +99 B |
| Scene 41 — Physics Shape Debug Viewer<br/>`scene41` | 92.64 KB | 92.80 KB | **167 B** | +7 B |
| Scene 267 — StandardMaterial Vertex Colors<br/>`scene267` | 41.49 KB | 41.70 KB | **217 B** | +7 B |
| Scene 43 — Parametric Proximity Path<br/>`scene43` | 57.71 KB | 57.93 KB | **222 B** | +8 B |
| Scene 18 — PCF Shadows<br/>`scene18` | 59.27 KB | 59.50 KB | **240 B** | +7 B |

…and 59 more, listed tightest first.

<details>
<summary>Tightest scenes repo-wide — 3 scenes over ceiling, 76 of 247 under 1.0 KB, 16 under 256 B</summary>

| Scene 163 - ShaderMaterial Alpha<br/>`scene163` ⬅ moved by this PR | 39.12 KB | 39.10 KB | ⚠️ 16 B over |
| Scene 159 - ShaderMaterial Basic<br/>`scene159` ⬅ moved by this PR | 39.38 KB | 39.38 KB | ⚠️ 3 B over |
| Scene 162 - ShaderMaterial Defines<br/>`scene162` ⬅ moved by this PR | 39.40 KB | 39.40 KB | ⚠️ 1 B over |
| Scene 161 - ShaderMaterial Uniform<br/>`scene161` ⬅ moved by this PR | 39.91 KB | 39.94 KB | 33 B |
| Scene 117 — 2D Sprite Picking<br/>`scene117` | 17.46 KB | 17.50 KB | 45 B |
| Scene 160 - ShaderMaterial Texture<br/>`scene160` ⬅ moved by this PR | 41.21 KB | 41.28 KB | 74 B |
| Scene 213 — GridMaterial<br/>`scene213` ⬅ moved by this PR | 49.43 KB | 49.52 KB | 94 B |
| Scene 16 — Thin Instances<br/>`scene16` ⬅ moved by this PR | 48.18 KB | 48.30 KB | 122 B |
| Scene 19 — PBR Clearcoat<br/>`scene19` ⬅ moved by this PR | 69.17 KB | 69.30 KB | 133 B |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` ⬅ moved by this PR | 44.66 KB | 44.80 KB | 143 B |

</details>

Popov72:
Thanks for your work on this PR. We're going to open a new PR incorporating these storage-buffer and GPU-resident geometry changes, adapted to current master, as a pre-integration step for compute shader support in Babylon Lite. The compute shader code will follow separately. The new PR will reference this one and preserve attribution for the original contributions.

We also plan to rework device-loss recovery in a dedicated future PR, so we're deferring further recovery-related changes to that work.

Popov72:
Storage-baked geometry for compute integration has been added in https://github.com/BabylonJS/Babylon-Lite/pull/737.

Compute shaders support has been added in https://github.com/BabylonJS/Babylon-Lite/pull/744.

Could you tell us if these PRs meet your needs?

VoxelBoy:
Yes — between #737 and #744 the storage and compute halves of what this PR proposed are covered, and in better shape than what I had. Closing this one as superseded.

I've checked the merged API against every engine change our fork carries for this work, rather than only the two scenes:

**Covered by #737** — `createStorageBuffer` with writable/vertex/index, `createMeshFromStorageBuffer` with slot addressing through `baseVertex`, shared topology, borrowed-buffer ownership (the fix for disposing one slot destroying the slab), explicit index format, and material-owned formats against mesh-owned packing. Our fork's `ShaderMaterialOptions.vertexLayout` is a different shape from `setShaderAttributeFormats`, but it covers the same ground; the split into declared format versus mesh packing is the better split.

**Covered by #744** — the one gap we had recorded against the compute seam is closed: our fork had no public seam for storage textures, so `cloudNoiseVolume.ts` fills a 3D noise volume by reaching into `engine._device` directly. `createComputeStorageTexture` with a `3d` view dimension and `depthOrArrayLayers` is exactly the seam that was missing. Binding sets per pipeline and one submission per task also replace the two things our fork had to add: `createComputeBindings` and `dispatchComputeBatch`, which existed because one binding set per program and one submission per dispatch cost us N pipelines and N submissions for N bodies.

Recovery being deferred costs us nothing. Manywhere treats device loss as terminal and visible: it deliberately does not enable scene recovery and rebuilds nothing, because a "successful" restore leaves every terrain mesh sourcing vertices from an orphaned allocation. So "applications must recreate affected resources after device loss" is stricter than what we ask for.

Two honest caveats. This is an API-level read plus the two parity scenes; the real test is porting Manywhere's terrain onto the merged API, which I haven't done yet. And our fork still carries a handful of patches that have nothing to do with these two PRs — `removeTask` as the inverse of `addTask`, a public `generateTextureMipmaps`, sampling options (address modes, filters, anisotropy) on `createRenderTargetTexture`, `mips` on `RenderTargetDescriptor`, and a per-draw index so one material can draw many objects. I'd rather raise those separately, each on its own merits, than bundle them here.

#584 is rebased onto master and is now just scene 286 and its golden — the visual coverage you kept out of #737. It passes at MAD 0.000 and measures 61.09 KB against a 61.36 KB ceiling, below the 63.53 KB it was originally authored with.

Thanks for taking the loader fix in so quickly, and for carrying the attribution through.

Issue #60, opened 2026-04-28
Large World Rendering


georginahalpern:
Owner: Georgie

Subproject for Large World Rendering (`useFloatingOrigin`) in Babylon Lite. See [`docs/architecture/27-large-world-rendering.md`](../blob/main/docs/architecture/27-large-world-rendering.md) for the full design. Milestones are feature-oriented — each one is independently shippable and lands with the tests that validate it.

## Milestones

- [x] **Milestone 1 — Foundations** (math + plumbing, no rendering changes yet)
  <details><summary>Details</summary>

  - Add `large-world/floating-origin.ts` with `getFloatingOriginOffset(scene) → Vec3` (zero when off); add `useFloatingOrigin?` to `SceneContextOptions` + `floatingOriginMode` on `SceneContextInternal`.
  - Compute and cache `scene._eyePosition` from active camera world matrix every frame.
  - Add Float64 variants of the matrix ops used to derive world matrices (`mat4.ts`: multiply, invert, lookAt, compose); wire `IWorldMatrixProvider` to optionally compute in F64 when the scene flag is on.
  - _Validation_: mirror `babylon.instancedMesh.lwr.test.ts` cases as pure-math unit tests (no GPU).
  </details>

- [x] **Milestone 2 — Mesh + thin-instance rendering** (first renderable; the 80% case)
  <details><summary>Details</summary>

  - In the per-mesh world-matrix uploader, subtract offset from `[12..14]` before writing to the GPU UBO when `floatingOriginMode` is on.
  - Zero translation in the `view` matrix when LWR is on (camera conceptually at origin).
  - Same recompose for `viewProjection` / `worldView` / `worldViewProjection` matrices the renderer uploads.
  - Subtract offset from `vEyePosition` uniform (PBR/Standard).
  - In `setThinInstances` upload path, subtract offset from translation columns (F64 source → F32 GPU).
  - Add the `TempFinalMat === mat` short-circuit guard so the offset isn't subtracted twice when an upload chain forwards the same matrix instance.
  - _Validation_: Lite scene `lab/sceneN.html` + BJS ref `lab/babylon-ref-sceneN.html` at ECEF-scale (camera ~1e6, mesh chain showing sub-unit precision survival; PG `#5U0N0Q` or `#P3E9YP#256`); golden capture (with user approval); parity spec with `maxMad` ceiling; bundle-size ceiling (verify scenes not importing LWR pay zero bytes); `pnpm test` green.
  </details>

[...]

PR #593, opened 2026-08-21
chore(large-world): drop the dead floating-origin updater hook and refresh its docs

sailro:
Follow-up to #60. When LWR moved off scene-side mirror state, the replacement landed but the old wiring did not get swept up.

`floating-origin.ts` states the change in its own header: the previous architecture kept `scene._floatingOriginOffset` / `_floatingOriginVersion` / `_eyePosition` in sync through a per-frame `updateFloatingOriginOffset`, and that was removed as *"net cost without value"*. Consumers now derive the offset from `scene.camera.worldMatrix` at the point of use.

What survived it:

**A dead hook on the engine type.** `engine.ts` still declared `_updateFOOffset?: (scene: SceneContext) => void`. Across tracked source in all three packages there were **zero assignments and zero call sites** — every remaining reference was the declaration or a comment:

```
$ git grep -n "_updateFOOffset\|updateFloatingOriginOffset" -- . ':!**/build/**'
docs/lite/architecture/35-large-world-rendering.md   (10 hits)
packages/babylon-lite/src/engine/engine.ts:163,168,374
packages/babylon-lite/src/large-world/floating-origin.ts:12
tests/lite/unit/floating-origin.test.ts:38,39
```

`updateFloatingOriginOffset` is no longer exported, and `_floatingOriginOffset` / `_floatingOriginVersion` / `_eyePosition` appear only inside comments, never as real fields.

**Two comments describing behaviour that isn't there.** `createEngine` attributed tree-shaking to a scene `_update` call on the hook — there is none. And `_wrapRenderableForFO` claimed it fires when `scene._floatingOriginVersion` changes, while the implementation tracks `scene.camera.worldMatrixVersion`. That second one documents a function still very much in use.

**The architecture page described the removed design throughout** — a "Per-frame offset update" section, code samples using fields that no longer exist, `camera._floatingOriginOffset` where the real gate is `camera._useFloatingOrigin`, and a tree-shaking proof resting on a call site in `scene-core.ts` that does not exist.

### What changed

- deleted `_updateFOOffset` and its doc block
- corrected the two engine comments and one stale test comment
- rewrote the drifted doc sections to the shipped design: `getFloatingOriginOffset`, `wrapRenderableForFO`, the real `createEngine` FO block, and a "Scene state: none" note replacing the removed field listing

The page keeps one deliberate reference to the old design — a short note recording that the mirror existed and why it went, mirroring the module header.

### Verification

The three code samples were checked mechanically against the source rather than by eye: extracted from both, comments stripped, whitespace collapsed, compared. `getFloatingOriginOffset`, `wrapRenderableForFO` and the `createEngine` FO block all match term for term.

`npx tsc --noEmit` clean, `npx eslint` clean, and the full unit project passes (217 files, 1666 tests).

### Risk

No runtime bytes: `_updateFOOffset` is an interface member, erased at build, so no manifest or ceiling movement.

I would expect the API report to be unaffected, since the member is `@internal` and underscore-prefixed and `report-api-changes.ts` trims `@internal` members. Flagging rather than assuming: if it happened to be the trailing internal member of `EngineContext`, that is precisely the formatting quirk the script's Prettier normalisation exists to absorb. I could not generate the report locally, so CI is the judge.

Happy to split the doc half out if you would rather own that page separately.

### DIFF
--- a/docs/lite/architecture/35-large-world-rendering.md
+++ b/docs/lite/architecture/35-large-world-rendering.md
@@ -75,17 +75,17 @@ export function getFloatingOriginOffset(scene: SceneContext): Vec3 {
 
 An earlier design kept `scene._floatingOriginOffset` / `_floatingOriginVersion` / `_eyePosition` in sync through a per-frame `updateFloatingOriginOffset` call. That was removed as net cost without value: the mirror had to be written every frame and read through an extra indirection, to hold a value the camera already carries. Invalidation now rides the camera's own `worldMatrixVersion` (see below).
 
-### Four places the offset is subtracted
+### Four places the offset is applied
 
-1. **`getViewMatrix(camera)`** (`camera/camera.ts`): when `camera._useFloatingOrigin` is set, the offset is subtracted from the camera world position _before_ the `R_inv * -cameraPos` calculation produces the view translation. When `offset == cameraPos` (the steady-state case), the resulting view translation is mathematically zero. The view matrix uploads therefore use the precision-only `packMat4IntoF32` (no 5th argument) — a second subtraction at upload would double-bias the translation.
+1. **`getViewMatrix(camera)`** (`camera/camera.ts`): when `camera._useFloatingOrigin` is set, the camera-position components are substituted with zero (`const cx = useFO ? 0 : w[12]!`, and likewise `cy`/`cz`) before the `-(R_inv * cameraPos)` calculation produces the view translation. No offset is subtracted — the offset _is_ the camera position, so under FO the view translation is exactly zero by construction. The view matrix therefore uploads through the precision-only `packMat4IntoF32`; routing it through the offset packer would bias the translation a second time.
 
-2. **Mesh-world UBO uploads** (`material/{standard,pbr,node}-renderable.ts`): each renderable's per-frame update calls `packMat4IntoF32(meshUboData, mesh.worldMatrix, 0, 0, _foOffset)`. The packer subtracts `_foOffset` from the translation column `[12..14]` during pack. Subtraction happens in JS number precision (F64) before the implicit F32 store, recovering the small remainder at full precision.
+2. **Mesh-world UBO uploads** (`material/{standard,pbr,node}-renderable.ts`): each renderable resolves its packer once at construction with `engine._makePackMeshWorld?.(scene) ?? packMat4IntoF32`, then invokes it with the same four arguments either way — `(view, mat, offsetFloats, srcOffsetFloats)`. Under FO that resolves to `makePackMeshWorld(scene)` (`large-world/pack-mat4-with-offset.ts`), whose `packMat4IntoF32WithOffset` subtracts the camera-derived offset from the translation column `[12..14]`. The subtraction happens in JS number precision (F64) before the implicit F32 store, recovering the small remainder at full precision.
 
-3. **`vEyePosition` uniform** (`frame-graph/render-task.ts`): `writePassSceneUBO` writes `camera.worldMatrix[12..14] - getFloatingOriginOffset(scene)` for the eye-position uniform. Shader expressions of the form `vEyePosition - input.worldPos` now produce the eye-relative vector at full precision because both sides live in the small-magnitude frame.
+3. **`vEyePosition` uniform** (`frame-graph/scene-uniforms-pack.ts`): `_packSceneUniforms` — which `_writePassSceneUBO` calls — writes literal zeroes into `data[32..34]` when `engine.useFloatingOrigin` is set, and the raw camera world position otherwise. It performs no subtraction and does not read the offset helper. Zero is the correct eye position because mesh world translations were already rebased by (2), so shader expressions of the form `vEyePosition - input.worldPos` produce the eye-relative vector at full precision with both sides in the small-magnitude frame.
 
 4. **ShaderMaterial system uniforms** (`material/shader/shader-renderable.ts`): ShaderMaterial writes its own UBO rather than going through `packMat4IntoF32`, so it cannot inherit the subtraction from (2). `_shaderWorldMatrix(mesh, camera, out?)` performs it instead, returning a camera-relative copy of `mesh.worldMatrix` that `world`, `worldView` and `worldViewProjection` are all derived from — they must share one frame, or passes reading different uniforms tear apart. `cameraPosition` is written as zero for the same reason `vEyePosition` is in (3). Both writers share the helper, so enabling `enable-shader-material-uniform-caching.ts` changes only how often the UBO is serialized, never what it contains. `LineMaterial` is a ShaderMaterial underneath and is covered by the same path.
 
-   This one keys on `camera._useFloatingOrigin` rather than the engine flag, so the test is bit-for-bit the one `getViewMatrix` applies to the *same* camera. A render-target task drawing through a non-scene camera then gets an untranslated view **and** an absolute world, which is self-consistent.
+    This one keys on `camera._useFloatingOrigin` rather than the engine flag, so the test is bit-for-bit the one `getViewMatrix` applies to the _same_ camera. A render-target task drawing through a non-scene camera then gets an untranslated view **and** an absolute world, which is self-consistent.
 
 ### Per-renderable version tracking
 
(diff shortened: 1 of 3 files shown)


packages/babylon-lite/src/engine/engine.ts
```diff
@@ -370,9 +361,9 @@ export async function createEngine(canvas: RenderCanvas, options?: EngineOptions
 
     // Same dynamic-import trick for the LWR runtime. When `useFloatingOrigin` is
     // false (the default) the `floating-origin.js` module is never referenced
-    // statically anywhere in the package — scene `_update` does
-    // `eng._updateFOOffset?.(scene)` which is a no-op when the field is
-    // undefined. Tree-shakers drop the module from non-LWR bundles.
+    // statically anywhere in the package: every consumer reaches it through an
+    // engine field left undefined when FO is off. Tree-shakers drop the module
+    // from non-LWR bundles.
```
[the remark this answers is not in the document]
sailro: Good catch, and correct — `index.ts:50` does `export { getFloatingOriginOffset } from "./large-world/floating-origin.js"`, so "never referenced statically anywhere in the package" is wrong as written.

For the record the wording predates this branch (it is on master in both `engine.ts` and the page), but I rewrote those sentences here, so they are mine to fix. Corrected in `b4bec2f`, in all three places — the engine comment and both doc paragraphs, including the line 126 instance you flagged.

Rather than soften it, the text now states the actual shape: the module has exactly **one** static edge, the package-root re-export of `getFloatingOriginOffset`; that edge is tree-shakeable and carries only the accessor, not the FO runtime. Everything else is reached through engine fields left undefined when FO is off, or through the `await import(...)` inside `if (useFO)`. So a non-LWR bundle contains neither the accessor nor the runtime — same conclusion, but a reader tracing why the chunk drops now gets the real reasoning instead of a claim they can falsify in one grep.

docs/lite/architecture/35-large-world-rendering.md
```diff
@@ -45,98 +45,73 @@ export function getFloatingOriginOffset(scene: SceneContext): Vec3;
...
+    ]);
+    _wrapRenderableForFO = wrapRenderableForFO;
+    _makePackMeshWorld = makePackMeshWorld;
+    _lightFoVersion = lightFoVersion;
+    _applyLightFoOffset = applyLightFoOffset;
 }
 ```
 
-The function reference is stored on `engine._updateFOOffset`. Scene `_update` invokes it via optional chaining:
+Each reference is stored on the engine (`engine._wrapRenderableForFO`, `_makePackMeshWorld`, `_lightFoVersion`, `_applyLightFoOffset`). Consumers reach them through optional chaining with a non-FO fallback, e.g. `engine._makePackMeshWorld?.(scene) ?? packMat4IntoF32`. Non-LWR engines leave every field undefined, so the LWR module is never referenced statically anywhere in the package. Tree-shakers drop it entirely from non-LWR bundles. Validated by `tests/parity/bundle-size.spec.ts` ceilings.
```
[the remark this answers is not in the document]
sailro: Good catch, and correct — `index.ts:50` does `export { getFloatingOriginOffset } from "./large-world/floating-origin.js"`, so "never referenced statically anywhere in the package" is wrong as written.

For the record the wording predates this branch (it is on master in both `engine.ts` and the page), but I rewrote those sentences here, so they are mine to fix. Corrected in `b4bec2f`, in all three places — the engine comment and both doc paragraphs, including the line 126 instance you flagged.

Rather than soften it, the text now states the actual shape: the module has exactly **one** static edge, the package-root re-export of `getFloatingOriginOffset`; that edge is tree-shakeable and carries only the accessor, not the FO runtime. Everything else is reached through engine fields left undefined when FO is off, or through the `await import(...)` inside `if (useFO)`. So a non-LWR bundle contains neither the accessor nor the runtime — same conclusion, but a reader tracing why the chunk drops now gets the real reasoning instead of a claim they can falsify in one grep.

docs/lite/architecture/35-large-world-rendering.md
```diff
@@ -45,98 +45,73 @@ export function getFloatingOriginOffset(scene: SceneContext): Vec3;
...
 
 2. **Mesh-world UBO uploads** (`material/{standard,pbr,node}-renderable.ts`): each renderable's per-frame update calls `packMat4IntoF32(meshUboData, mesh.worldMatrix, 0, 0, _foOffset)`. The packer subtracts `_foOffset` from the translation column `[12..14]` during pack. Subtraction happens in JS number precision (F64) before the implicit F32 store, recovering the small remainder at full precision.
 
-3. **`vEyePosition` uniform** (`frame-graph/render-task.ts`): `writePassSceneUBO` writes `camera.worldMatrix[12..14] - scene._floatingOriginOffset[0..2]` for the eye-position uniform. Shader expressions of the form `vEyePosition - input.worldPos` now produce the eye-relative vector at full precision because both sides live in the small-magnitude frame.
+3. **`vEyePosition` uniform** (`frame-graph/render-task.ts`): `writePassSceneUBO` writes `camera.worldMatrix[12..14] - getFloatingOriginOffset(scene)` for the eye-position uniform. Shader expressions of the form `vEyePosition - input.worldPos` now produce the eye-relative vector at full precision because both sides live in the small-magnitude frame.
```
VicenteCartas: This is not the shipped behavior. `_writePassSceneUBO` delegates to `_packSceneUniforms` in `frame-graph/scene-uniforms-pack.ts`; that function writes literal zeroes to `data[32..34]` when `eng.useFloatingOrigin` is enabled, and otherwise writes the raw camera world position. It neither calls `getFloatingOriginOffset` nor performs this subtraction. The current wording also contradicts the later “exactly one static edge” statement, because such a call from the frame graph would add another static edge.

Please describe `_packSceneUniforms` writing `vEyePosition = (0, 0, 0)` under FO, name `scene-uniforms-pack.ts` as the implementation site, and treat `_writePassSceneUBO` as its caller.

docs/lite/architecture/35-large-world-rendering.md
```diff
@@ -45,98 +45,73 @@ export function getFloatingOriginOffset(scene: SceneContext): Vec3;
...
-    /** Camera world position copied each frame by
-     *  `updateFloatingOriginOffset` so consumers can read eye position
-     *  without bouncing through the camera. */
-    _eyePosition: [number, number, number];
-}
-```
+### Scene state
 
-The fields exist on every scene regardless of `useFloatingOrigin` to keep the renderable closure shape uniform. For non-LWR scenes the version stays 0 and the offset stays `[0, 0, 0]`, so all subtractions are no-ops and all version comparisons are dead.
+None. LWR adds no fields to `SceneContext`: the offset is the active camera's world position, read on demand, and invalidation rides that camera's `worldMatrixVersion`. Non-LWR scenes therefore carry no LWR state at all, rather than inert zeroed fields.
```
VicenteCartas: “No `SceneContext` fields” is accurate, but removing `scene/scene-core.ts` from the page drops surviving, required LWR lifecycle behavior. `scene._update` is still the sole writer of `camera._useFloatingOrigin = true` and invalidates `_viewVer` / `_vpVer`; `getViewMatrix` depends on that flag.

Please restore a `scene/scene-core.ts` Files / size row describing that camera-flag and view-cache wiring, and scope this paragraph explicitly to the absence of stored `SceneContext` mirror fields.

docs/lite/architecture/35-large-world-rendering.md
```diff
@@ -45,98 +45,73 @@ export function getFloatingOriginOffset(scene: SceneContext): Vec3;
...
 
 ### Four places the offset is subtracted
 
-1. **`getViewMatrix(camera)`** (`camera/camera.ts`): when `camera._floatingOriginOffset` is set, the offset is subtracted from the camera world position _before_ the `R_inv * -cameraPos` calculation produces the view translation. When `offset == cameraPos` (the steady-state case), the resulting view translation is mathematically zero. The view matrix uploads therefore use the precision-only `packMat4IntoF32` (no 5th argument) — a second subtraction at upload would double-bias the translation.
+1. **`getViewMatrix(camera)`** (`camera/camera.ts`): when `camera._useFloatingOrigin` is set, the offset is subtracted from the camera world position _before_ the `R_inv * -cameraPos` calculation produces the view translation. When `offset == cameraPos` (the steady-state case), the resulting view translation is mathematically zero. The view matrix uploads therefore use the precision-only `packMat4IntoF32` (no 5th argument) — a second subtraction at upload would double-bias the translation.
```
VicenteCartas: The field-name correction is right, but this explanation still uses the removed mirror model. `getViewMatrix` does not subtract an offset or rely on an `offset == cameraPos` steady state; when `_useFloatingOrigin` is set, it directly substitutes zero for the camera-position components before constructing the view translation.

The same stale design remains elsewhere in this edited section and should be swept in this PR:

- item 2 shows a nonexistent five-argument `packMat4IntoF32(..., _foOffset)` call; renderables select `engine._makePackMeshWorld?.(scene) ?? packMat4IntoF32` and invoke the selected packer with four arguments;
- the Validation bullets still claim `floating-origin.test.ts` covers the removed per-frame updater/version bumps and call the offset helper `packMat4IntoF32`;
- the Files / size table assigns an `offsetXYZ` fifth argument to `math/pack-mat4-into-f32.ts`, but the offset implementation is `large-world/pack-mat4-with-offset.ts` / `packMat4IntoF32WithOffset`.

Please update these claims together so the page consistently describes the shipped design.

VicenteCartas (changes requested):
The dead `_updateFOOffset` removal itself looks correct and safe: I found no assignments, calls, dynamic access, or remaining API/runtime dependency, and the engine/test comment changes match the implementation. The refreshed architecture page still mixes in several parts of the removed design, though, so I am requesting changes for documentation accuracy.

The blocking details are inline. In short:

- document the real `vEyePosition` writer and value under floating origin;
- retain the surviving `scene-core.ts` camera-flag/cache-invalidation lifecycle;
- finish the sweep of old subtraction, packer-signature, and test-coverage claims.

Non-blocking: `npx prettier --check docs/lite/architecture/35-large-world-rendering.md` fails. Comparing base and head shows the PR adds a table-alignment formatting hunk around the Files / size table, so please run Prettier on the document.

Focused validation at this exact head: Babylon Lite and test TypeScript checks passed, scoped ESLint passed, and 34/34 focused floating-origin/renderable tests passed. No runtime bundle change is expected from deleting an erased internal interface member.

VicenteCartas (changes requested):
Left some minor review items.

docs/lite/architecture/35-large-world-rendering.md
sailro: Addressed in `70b2ac5`. You were right — I wrote that subtraction from the shape of the surrounding text instead of reading the writer, and it was wrong twice over: `_packSceneUniforms` (`frame-graph/scene-uniforms-pack.ts`) writes literal zeroes into `data[32..34]` when `engine.useFloatingOrigin`, otherwise the raw camera world position, and it never touches the offset helper — which, as you note, would also have contradicted the one-static-edge claim.

The item now names `scene-uniforms-pack.ts` as the implementation site with `_writePassSceneUBO` as its caller, states the zero write, and explains why zero is the correct eye position: mesh world translations were already rebased by site (2), so both sides of `vEyePosition - input.worldPos` are in the same small-magnitude frame.

docs/lite/architecture/35-large-world-rendering.md
sailro: Addressed in `70b2ac5`. Dropping the `scene-core.ts` row was an over-correction: I scoped the section to "no `SceneContext` fields", which is true, and in doing so removed live behaviour.

The Files / size table now carries a `scene/scene-core.ts` (`_update` FO block) row for stamping `camera._useFloatingOrigin` and invalidating `_viewVer` / `_vpVer`, and the paragraph is scoped to the absence of *stored mirror fields* rather than implying no LWR wiring survives. It also quotes the four-line `_update` block verbatim, so the flag `getViewMatrix` depends on is visible on the page and its sole writer is named.

docs/lite/architecture/35-large-world-rendering.md
sailro: Addressed in `70b2ac5`, all four points.

- `getViewMatrix`: rewritten to the substitution model — `const cx = useFO ? 0 : w[12]!` and likewise `cy`/`cz`, so the view translation is exactly zero by construction. The "offset == cameraPos steady state" argument is gone; as you say, the offset *is* the camera position, so there was nothing being subtracted.
- Item 2: now `engine._makePackMeshWorld?.(scene) ?? packMat4IntoF32`, resolved once at construction and invoked with the same four arguments either way, with `packMat4IntoF32WithOffset` named as the FO packer. Confirmed `packMat4IntoF32` has no fifth parameter.
- Validation bullets: rewritten to the cases the suites actually run — `getFloatingOriginOffset` returning the camera position and zero without a camera, `scene._update` stamping the flag only under FO, and in the upload suite `getViewMatrix` zeroing with its FO-off control, `packMat4IntoF32WithOffset` landing `delta` at world `1e6 + delta`, and the eye-relative-zero write.
- Files / size: the `offsetXYZ` row is gone; `math/pack-mat4-into-f32.ts` is described as the precision-only packer, with `large-world/pack-mat4-with-offset.ts` carrying the offset variant.

Section heading changed from "subtracted" to "applied", since sites (1) and (3) substitute zero rather than subtract.

Also fixed your non-blocking point: `npx prettier --check` passes on the page now.

Every code sample on the page is checked against its source mechanically — extracted from both, comments stripped, whitespace collapsed, compared — and all four match. That check is what I should have been running on the prose claims too.

VicenteCartas (approved):
Comments addressed, thanks!

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/593/)

<sub>Build 20260825.43 - merge @ 03b5757340b2c92d0bf1d07ca647ec3bcbc2b229</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.43/lab/index.html)

<sub>Build 20260825.43 - merge @ 03b5757340b2c92d0bf1d07ca647ec3bcbc2b229</sub>

PR #46, opened 2026-04-22
refactor(material): lazy-init PBR/Standard extension registries

deltakosh:
Module-level
ew Map() allocations defeat tree-shaking because bundlers treat them as side-effectful. Both _pbrExts (pbr-flags.ts) and _stdExts`n(standard-pipeline.ts) now use the lazy-init pattern blessed by GUIDANCE.md §4.

This keeps scenes that never touch PBR/Standard extension machinery from paying for registry scaffolding at import time.

PR #611, opened 2026-08-25
fix(device-lost): rebuild every captured texture, and refuse to abandon contexts

ryantrem:
Two independent use-after-frees in WebGPU device-lost recovery. Both let recovery report success while a GPU object was still bound to the lost device, so nothing fails cleanly — the next touch kills the browser's renderer process (`STATUS_ACCESS_VIOLATION` / "Aw, Snap") with a stack pointing nowhere near device loss, and Dawn backend validation hides both.

Found by a downstream text-layout app whose page died ~1 s *after* lite reported a successful recovery.

> [!IMPORTANT]
> **This PR raises a bundle-size ceiling and needs an explicit sign-off before merge.** See [Bundle size](#bundle-size) below. @deltakosh — could you confirm the scene164 `maxRawKB` bump is acceptable, or tell me to find the bytes elsewhere?

## 1. Textures no rendering context references were never rebuilt

Recovery discovered textures purely by walking registered contexts' object graphs (scene materials, sprite layers, text GPU state). A recoverable texture the app owns *outside* those graphs therefore kept its dead-device `GPUTexture` straight through a "successful" recovery.

The trigger is a **transient dereference**, not "never used":

1. app renders in bitmap mode → a sprite atlas page is populated
2. app switches to outline mode → that page is no longer referenced by any `Sprite2DLayer`
3. device is lost → the reachability walk cannot see the page, so it is skipped
4. app switches back to bitmap → the stale texture goes into `createBindGroup` on the new device → renderer process dies

Instrumentation from the reporting app caught it exactly:

```
[PROV] STALE GPUQueue.writeTexture gen=1 current=2
    at appendSpriteAtlasFrames → appendBitmapAtlasFrame → drawGlyphRunBitmap
```

Capture already keeps a byte-accurate CPU shadow of precisely these textures, so everything needed to restore them was present — only the *discovery* was missing.

**Fix.** Every recovery source now flows through one `stampTexture(state, tex, source)` seam in `device-lost-recovery-capture.ts` that assigns `_recoverySource` *and* weakly tracks the texture. Recovery rebuilds that tracked set (`rebuildRecoverableTextures`) after surfaces are reconfigured but **before** any per-kind handler runs, so handlers can bind rebuilt textures.

The existing reachability walks stay — they are still how a texture that was never captured through `_dlr` gets reached.

A per-device idempotency guard in `rebuildTexture2D`, keyed on the recovery **source**, keeps the walks from rebuilding a second `GPUTexture`. That guard also closes a **pre-existing latent bug**: scene and sprite recovery use separate visited `Set`s, so a texture shared between a material and a sprite layer was already being rebuilt twice — orphaning the first `GPUTexture` and re-fetching `url` sources over the network. It is keyed by device rather than a boolean so a *second* device loss still rebuilds, and is recorded synchronously before any `await`.

### 1b. Wrappers derived by spreading a base — review follow-up

@RaananW caught that stamping alone was **not** complete. `cloneTexture2D(base, transform)` is `{ ...base, ...transform }`, and the glTF sampler path does `{ ...tex, sampler }`. Both inherit `_recoverySource` without ever passing through the stamp, and both produce a *distinct* wrapper owning its own `texture` field. Reading such a wrapper after recovery hits the same use-after-free, one hop removed.

A follow-up review found the sharper version of it: a clone does not keep its base alive. If the app holds the clone and drops the base, there is no live tracked wrapper for that source at all, so nothing ever initiates its rebuild.

**Fix.** Derived wrappers are tracked as captured textures in their own right.

`cloneTexture2D` notifies a module-level hook that the capture module installs. A hook is needed because neither `cloneTexture2D` nor the glTF `wrapTexture`/`wrapTexCoord` hooks that call it are handed an `engine`, and because a scene that never enables recovery should carry none of the bookkeeping. The capture module records which recovery state captured each source and adds the derived wrapper to that engine's tracked set. That state is held through a `WeakRef`, so an app that keeps a texture past its engine does not pin the engine's registrations and, through them, its whole scene graph.

Tracking them in the **existing** set rather than a parallel registry is what makes the base-collected case work, and it means derived wrappers are compacted on the same schedule as every other tracked texture.

Because rebuilding is deduped on the shared **source**, a whole wrapper family still costs a single upload and a `url` source is not re-fetched once per wrapper. A wrapper reached while that rebuild is still in flight awaits it instead of adopting the lost device's texture. Adoption resolves each wrapper's own captured sampler descriptor, so the glTF sampler wrapper keeps its own wrap/filter settings rather than silently inheriting the base's.

### 1c. Ownership of rebuilt textures — review follow-up

@deltakosh caught that rebuilding a texture was not the same as restoring it.

`createTexture2D`, `createTexture2DFromPixels` and `createRenderTexture2D` each `acquireTexture` the texture they hand back. That creator-owned reference is what keeps a texture alive for as long as the app holds the wrapper, independently of whoever is currently binding it. A replacement `GPUTexture` starts at ref-count **zero**, so a rebuilt `url`, `pixels` or `render` texture was destroyed by the first consumer that bound and then unbound it:

```
acquireTexture(tex)  // material binds  → 0 + 1 = 1
releaseTexture(tex)  // material unbinds → 1 - 1 = 0 → destroy()
```

— a use-after-free that recovery introduced itself, on a texture the app still held. Recovery now carries that ownership onto the replacement.

The other direction was wrong too: a texture the app had already released was still rebuilt, allocating a replacement and taking a reference nothing would ever release. Recovery now skips a released texture.

### 1d. Ownership, corrected twice more — review follow-up

@deltakosh followed up that those two questions were still tangled: the release check reused the *re-acquire* predicate (`url`/`pixels`/`render`), and they do not have the same answer.

**`dynamic` was a real miss.** It is creator-owned too (`dynamic-texture.ts:128`), but it was excluded from re-acquire for an unrelated reason — its own rebuild module restores its reference (`dynamic-texture-recovery.ts:78`), so acquiring again would double it. Sharing one predicate silently dragged that exclusion into the release check, and a released dynamic texture was rebuilt.

**Solid and glTF-uploaded textures can reach zero as well** — `releaseTexture` is public API and its `(m.get(t) ?? 1) - 1` default means the first release on a never-acquired texture destroys it. But they could not simply be added to the guard: with a two-state map their destroyed state is *indistinguishable* from their normal live state, since their creators never acquire. Treating "no owner" as released fails 3 tests including both glTF sampler recovery tests — every solid and glTF texture in a scene would be skipped and left on the lost device.

So the map now distinguishes three states: **absent** (nothing ever took ownership), **positive** (owned), **zero** (every owner released it and it was destroyed). `releaseTexture` leaves the zero behind rather than deleting it; the key is weak either way, so it retains nothing. The guard is now kind-independent and covers all six kinds, while re-acquisition keeps its own separate predicate.

**And a second ownership bug surfaced next to it.** Recovery restored exactly one reference, but `cloneTexture2D` documents that acquire/release pairing is the caller's, so a derived family can hold several references to one `GPUTexture`. A family at 2 came back at 1; the app then made the two releases it correctly owed, and the first destroyed the texture while the clone still pointed at it. Recovery now reads the outgoing texture's owner count before the rebuild replaces the handle and tops the replacement up to it. Counting instead of re-acquiring also deletes the kind list: `dynamic`'s self-restore is counted rather than doubled, and kinds that never acquired have nothing to carry.

### 1e. That ownership top-up ran too early — review follow-up

@deltakosh caught that carrying the count *at rebuild time* double-counts every reference recovery re-establishes itself: `rebuildSceneGpu` discards a mesh's queued texture releases and re-acquires as it rebuilds bind groups, and the dynamic-texture rebuild re-takes its own reference. A url texture held by its creator and one mesh went 2 → restored to 2 → 3 once the scene handler ran, and rested at 1 after every release the app would ever make. It was therefore never destroyed, and `_isTextureReleased` never reported it released — disabling the detection in 1d on any later loss.

**Fix.** `rebuildTexture2D` queues the ownership instead of carrying it, and `runDeviceLostRecovery` settles the queue once every handler in `_recoverOrder` has run. Whatever is still missing at that point is by definition a reference recovery did not re-establish, so there is no provenance to track and no ordering to get right. It stays a top-up rather than a flat acquire, so a handler that already re-took its reference is counted while a derived clone family — which no handler walks — still returns to 2.

Settling the abandoned references *before* the snapshot was tried and rejected: a solid or bitmap texture bound by one renderable sits at 1, draining it early writes the zero entry that marks it destroyed, and `rebuildTexture2D` then skips the rebuild entirely — leaving the dead texture for the re-acquired bind group. That is a worse use-after-free than the inflation it fixes.

### 1f. The deferred queue was shared by every engine — review follow-up

@deltakosh caught that the queue 1e added was module-level, so every engine and every run shared it. That matters because a lost GPU process loses **every device on the page at once** — two engines recovering concurrently is the ordinary case, not an exotic one. Both of their async rebuilds append to the one queue, and whichever run reaches its `finally` first drains both: the other engine is topped up to its full pre-loss count while its handlers have yet to re-establish anything, those handlers then acquire on top, and the texture rests one reference over. Never destroyed, and `_isTextureReleased` never reports it released — 1e's inflation, moved from within one engine to across two. The `catch` in `rebuildRecoverableTextures` had the same reach: one engine's rebuild failure settled the other's queue.

**Fix.** The queue lives on that engine's `DeviceLostRecoveryState` and the settle callback closes over it. Per-engine is only *also* per-run once the coordinator enforces it, which it did not — see 1g.

It is reached through the engine rather than threaded as a parameter because two of `rebuildTexture2D`'s three call sites — `recovery-rebuild.ts` and `sprite-recovery.ts` — are the handlers' own walks, running *during* the handler phase. A parameter those sites did not pass would silently drop the ownership of whatever they rebuilt, a quieter version of the same bug. Reading it off the engine keeps every caller correct with no signature churn, matching how `recoverCapturedSampler` already reaches `engine._deviceLostRecovery?._samplerDescriptors`.

Costs nothing on the always-bundled path — the field is optional and `getState` never initializes it, the new import is `import type` and erases, and `texture-recovery.ts` is a lazy chunk. All six tightest scenes measure byte-identical to the previous commit.

### 1g. An engine could overlap itself after all — review follow-up

@deltakosh caught that 1f's per-engine queue rested on an invariant the coordinator never enforced. `arm()` is reached from `_enableDeviceLostRecovery` as well as from a resolved run, and a run assigns `engine._device = D2` long before its handlers finish while `_armedDevice` still names `D1`. So a registration enabled in that window finds `_armedDevice !== engine._device`, arms `D2`, and a loss of `D2` starts a **second run on the same engine**. A crash-looping GPU process produces that back-to-back loss without any exotic setup. The two runs then share one `_pendingOwnership` queue — 1f's bug, reintroduced within a single engine — and, as he noted, also contend for that engine's `_device` and surface list, so scoping the queue alone was never sufficient.

The comment 1f shipped ("an engine cannot overlap itself") described a coincidence of one code path, not an invariant, and new shared state was then built on top of it.

**Fix.** `_recovering` on the recovery state; `arm()` refuses while a run is in flight. Set before the `_onLost` dispatch, cleared in both continuations:

- **Success** clears, then re-arms. Nothing is dropped by having waited — `GPUDevice.lost` is a promise, so a device lost during the window resolves for that later subscriber too and the deferred recovery runs immediately afterwards, serialized behind the run already in flight.
- **Failure** clears without re-arming. A failed run can leave the engine on an already-lost device, and arming it would spin recovery forever. Terminal-failure behaviour is unchanged; the clear only restores the ability of a later `_enableDeviceLostRecovery` to arm, which it had before 1g.

Keying the queue by device instead was considered and rejected: it makes the queue safe while leaving two runs fighting over one engine's device and surfaces.

The field is optional so `getState` is untouched, leaving the guard and three assignments as the whole always-bundled cost — **+30 B on scene164**, the only lab scene that enables device-lost recovery, which keeps 194 B of headroom. The other five tightest scenes are byte-identical.

## 2. Registered contexts of kinds with no enabled strategy were left behind

`runDeviceLostRecovery` ran `_recover` only for kinds with a registered strategy, then reported success and re-armed. Every *other* registered rendering context kept its buffers, bind groups and pipelines from the dead device, and the next `renderFrame` encoded draws against freed native objects.

The realistic shape is an app that enables sprite recovery, has a scene context registered, and never enables scene recovery.

**Fix.** Recovery now fails up front, after `stopEngine` but *before* anything is disposed or a replacement adapter is requested — so `onRecoveryFailed` fires with nothing torn down and the app learns this engine has to be discarded, rather than silently continuing toward a renderer crash.

This is a behaviour change, but only for a configuration whose previous behaviour was "report success, then crash the browser". It turns a silent UAF into an actionable error naming the offending kinds.

No interaction with the `xr-session` kind from #537 — that registers *with* a `_recover` handler, so the guard accepts it.

## Verification

End-to-end against the original real-world crash, hardware adapter, GPU process killed via CDP `Browser.crashGpuProcess`:

| arm | reps | crashes |
|---|---|---|
| unpatched control | 3 | **3/3 crash** (`STALE GPUDevice.createBindGroup gen=1 current=2`) |
| this fix | 4 | **0/4** |

The downstream app needs **zero** code changes: `rebuildTexture2D` mutates the wrapper in place (`texture`/`view`/`sampler`/`width`/`height`) and never touches `frames`/`_packState`, so existing references stay valid.

Reproduced independently in-repo as well — 2/2 crash with the texture rebuild disabled, 0/2 with it enabled.

That verification predates the review follow-ups in 1b through 1g, which are covered by unit tests rather than another hardware run.

## Tests

`tests/lite/unit/device-lost-recovery.test.ts` grows to 42 tests:

- 4 on the unreferenced-texture rebuild — tracking, the idempotency guard, rebuild-before-handlers ordering, and re-rebuild after a second loss
- 5 on derived wrappers — adopting the base's upload, the **base-collected** case, per-wrapper sampler recovery, one-upload-per-source when a walk reaches a clone whose base was already rebuilt, and a single shared rebuild leaving a base and two clones on the same texture, view *and* sampler
- 12 on released-texture detection — a 6-kind × 2-scenario table (a single wrapper, and a family whose clone performs the final release), replacing the two hand-written cases
- 9 on ownership — each creator-owned kind surviving a consumer acquire/release cycle, an ownerless texture nothing released still being rebuilt, a derived family's second reference surviving the rebuild, `dynamic` not being double-acquired, a texture whose handler re-acquires it still reaching zero and being destroyed on final disposal, and two engines whose recoveries interleave — one parked inside its handlers while the other completes — settling only their own textures
- 3 on the unrecoverable-kind guard — message contents, dedupe + stable sort, and that it throws before anything is disposed
- 1 on coordinator serialization — driven through the real `arm()` path rather than by calling `runDeviceLostRecovery` directly: lose a device, park the run inside its handlers, enable a further recovery kind and lose the replacement, then assert no second run starts on that engine while the first is in flight *and* that the deferred loss is still recovered once it settles

Every one of these was checked against buggy code rather than just written green:

| mutation | fails |
|---|---|
| tracking hook stubbed out | 3 of the 4 derived tests |
| restore the old kind-limited release guard | 6 — exactly `dynamic`/`solid`/`bitmap` |
| `delete` instead of `set(t, 0)` on destroy | 12 — every skip test |
| treat "never owned" as released | 3 — including both glTF sampler tests |
| creator-only re-acquire | 1 — the clone family test |
| flat re-acquire for every kind | 2 — clone family + dynamic |
| carry ownership at rebuild time instead of settling | 1 — the handler re-acquisition test |
| flat `acquireTexture` in the settle instead of a top-up | 2 — clone family + dynamic |
| share one module-level queue across engines | 1 — the interleaved two-engine test, on the release behaviour alone with its bookkeeping assertions disabled |
| drop the `_recovering` guard from `arm()` | 1 — the coordinator serialization test, `expected 1 times, got 2` |

On that last row, one caveat stated rather than left to be discovered: with the two direct assertions also disabled, the ref-count assertions did **not** catch the overlap on their own — in that interleaving the two runs' ownership math still balances. The test's comment now says so, and the call-count and device-identity assertions are named as the actual detector.

`tests/lite/unit/dynamic-texture.test.ts`'s `_dlr` stub now performs the assignment, mirroring the real stamp contract.

## Docs

`docs/lite/architecture/50-device-lost-recovery.md` is updated for the fail-fast behaviour on unrecoverable kinds (replacing the obsolete "context kinds without a registered handler are left untouched" contract and its matching test requirement), the up-front rebuild of captured textures, derived wrappers, and the ownership rules above.

## Bundle size

**⚠️ This raises a ceiling and needs a maintainer's explicit approval — please do not merge on my say-so.**

scene164's `maxRawKB` goes **141.7 → 144**. The scene measures **143.5 KB**, under the proposed ceiling. It is the only lab scene that calls `enableDeviceLost*`.

**No other ceiling changes.**

### An always-bundled regression, found by CI and fixed

The first bundle run failed five unrelated scenes by 8–48 bytes. I had measured scene50 and scene92 as byte-identical and concluded the always-bundled cost was zero — wrong, because neither scene loads the affected modules. The real cause was structural, not line count:

- `gltf-sampler-desc` gained a **runtime** import of `texture-2d` where it previously imported only types. That edge pulls `texture-2d` into bundles that never load it at runtime. scene248 gained a **92-byte `texture-2d` chunk it did not have at all**, plus 63 bytes in `gltf-sampler-desc` — 173 bytes against 140 bytes of headroom.
- The released-texture check sat in `releaseTexture`, so every scene paid for it whether or not it ever recovers.

Both are now off the always-bundled path:

- `gltf-sampler-desc` has an engine in scope, so it registers its derived wrapper through the existing `engine._dlr` capture seam — a property access, no module dependency — and keeps its type-only import. `cloneTexture2D` has no engine (public API, reached from glTF through `GltfFeature.wrapTexture`), so it keeps the module-level hook, now inlined to drop the extra exported indirection. Both call the same tracking function, and behaviour is unchanged.
- Recovery reads the owner count through a tree-shakeable `_textureRefCount` instead of `releaseTexture` clearing `_recoverySource`. For the three creator-owned kinds a count of zero means the last release already destroyed the texture, which is exactly what the flag stood for. Kinds whose creator never acquires are exempt, since zero is their steady state.

Measured against `master`, the five scenes that failed:

| scene | before | after | headroom on master |
|---|---|---|---|
| scene12 | +69 B | **+1 B** | 28 B |
| scene27 | +66 B | **+15 B** | 43 B |
| scene73 | +70 B | **+1 B** | 62 B |
| scene248 | +173 B | **+38 B** | 140 B |
| scene253 | +134 B | **+50 B** | 86 B |

The `texture-2d` chunk delta is gone entirely; what is left is the derived-wrapper registration in `gltf-sampler-desc` itself.

Two notes for anyone reproducing this. Measure the **merge** with `master`, not the branch alone — Azure validates the merge commit, and branch-only numbers do not match it. And do not set `CI=1` locally: it forces SwiftShader, which cannot re-acquire an adapter after scene164's deliberate device loss and fails with `#46` instead of a size result.

## Known gaps, deliberately not fixed here

- **Textures with no recovery source at all.** `createRenderTargetTexture`, `createTexture2DArray`/`FromKtx2`, `createTexture3DFromPixels` and `createHtmlTexture` are all public API and none of them stamp a source, so recovery still cannot rebuild them. Tracking them would be dead weight without a source to rebuild *from*; giving them sources is a larger design change.
- **`ensureSharedAtlasGpu` calls `destroy()` across devices.** `text/_gpu/text-textures.ts` destroys `curveTex`/`bandTex` when `gpu.device !== device`. Pre-existing, spec-legal, and inert in 7/7 verification reps, but it should skip the destroy when the device is lost.
- **No public "is the device lost?" predicate.** Every consumer hand-rolls callback latching to answer it.

## Validation

- `pnpm test:unit` — 247 files, 2640 tests pass (3 skipped)
- `pnpm lint` — eslint + all 6 typecheck projects clean
- `pnpm build:bundle-scenes` — scene164 143.5 KB, plus scene12/27/73/248/253 all back under their unchanged ceilings

scene164's parity spec (MAD was 0.0000) is left to CI.

<details>
<summary>Merge notes</summary>

Merged `master` twice. The second merge picked up #526, which untracks the ~227 generated bundle manifests, so this PR's original `lab/public/bundle/manifest/scene164.json` edit was dropped — the ceiling in `scene-config.json` is now the only bundle-size change here.

While validating that merge locally, three specs failed (`bundle-content-no-f64`, `shadow-tracking-bundle-isolation`) purely because they read manifests that no longer ship in git and I had no local build. They pass after `pnpm build:bundle-scenes`, and CI builds before running them, so this is not a regression.

Small latent nit for whoever owns #526, unrelated to this PR: those specs guard with `it.skipIf(!existsSync(BUNDLE_DIR/sceneN.js))` but then read `BUNDLE_DIR/manifest/sceneN.json`. With a stale `sceneN.js` on disk and no manifest, they hard-fail instead of skipping. Keying the guard on the manifest would make the skip honest.

</details>

<details>
<summary>Note for anyone measuring scene164 locally</summary>

`measurementBrowserArgs()` switches the measurement browser to SwiftShader when `process.env.CI` is set. scene164 deliberately destroys its device, and its post-loss `requestAdapter()` then returns null, so the run fails with `#46 "WebGPU adapter not available"` rather than a size regression. If you export `CI=1` in your shell for unrelated reasons, unset it before measuring this scene.

</details>


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.1 - merge @ 90838c657e8ae2186e004f885ce0d062080e6166</sub>

tests/lite/unit/xr/xr-session.test.ts
```diff
@@ -587,6 +587,8 @@ describe("xr-session lifecycle", () => {
             _samplerDescriptors: new WeakMap(),
             _captureRefs: 0,
             _meshCaptureRefs: 0,
+            _textures: new Set(),
+            _texturesPruneAt: 0,
         };
```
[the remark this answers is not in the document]
ryantrem: Good catch — fixed in bc22e206, now seeded to `64` to match the coordinator.

Worth noting for anyone reading later that this was latent rather than active: with `0`, the first stamp takes the prune branch (`0 >= 0`), scans an empty set, and then sets `_texturesPruneAt = Math.max(64, 0 * 2)` — so it self-corrects to the production value after one stamp. And this particular test stamps nothing at all. But you're right that a fixture disagreeing with production defaults is exactly the kind of thing that quietly masks a regression later, so it shouldn't stay that way.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.3 - merge @ 73342ee872d4ed55f1e1d685c97d8874b0e49839</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 164 - Device Lost Recovery<br/>`scene164` | 143 KB | 142 KB | **+1 KB** |

### Decreases

| Scene 4 — Shadows<br/>`scene4` | 71 KB | 72 KB | -1 KB |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.3/lab/index.html)

<sub>Build 20260825.3 - merge @ 73342ee872d4ed55f1e1d685c97d8874b0e49839</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.14 - merge @ 9b13b03f9544315f85b887f1b22e3b8a013b0f21</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.14/lab/index.html)

<sub>Build 20260825.14 - merge @ 9b13b03f9544315f85b887f1b22e3b8a013b0f21</sub>

packages/babylon-lite/src/engine/device-lost-recovery-capture.ts
```diff
@@ -2,37 +2,61 @@ import type { EngineContext } from "./engine.js";
...
+        const textures = state._textures;
+        if (textures.size >= state._texturesPruneAt) {
+            for (const ref of textures) {
+                if (!ref.deref()) {
+                    textures.delete(ref);
+                }
+            }
+            state._texturesPruneAt = Math.max(TEXTURE_PRUNE_FLOOR, textures.size * 2);
+        }
+        textures.add(new WeakRef(tex));
```
RaananW: This tracks only the wrapper passed to `stamp`. `cloneTexture2D` and the glTF sampler path create new wrappers by spreading the original, so they inherit `_recoverySource` but are never added here. If such a wrapper is outside a registered context during device loss, recovery updates only the original wrapper and the clone retains the lost-device `GPUTexture`; using it later can reproduce the renderer crash. Could derived wrappers share this weak tracking and be updated during recovery too?

deltakosh:
Scene 164 bundle size raise is approved

deltakosh (changes requested):
Restore the creator-owned reference when URL, pixels, and render textures are rebuilt. Add coverage that each replacement survives one consumer acquire and release cycle.

Do not rebuild textures that were already released or disposed. Add coverage for a released captured texture.

Keep `cloneTexture2D` wrappers linked to their base through recovery. Rebuild the shared GPU resources once, and make out-of-graph clones adopt the base's rebuilt texture, view, and sampler. Add coverage for a single rebuild with preserved sharing.

Update `50-device-lost-recovery.md` for the new fail-fast behavior and the up-front rebuild of captured textures. Replace the obsolete skipped-unregistered-kinds test requirement.

packages/babylon-lite/src/engine/device-lost-recovery-capture.ts
ryantrem: Good catch — you were right, and the gap was slightly worse than it first looked.

There are exactly two spread sites: `cloneTexture2D`'s `{ ...base, ...transform }` and `gltf-sampler-desc.ts`'s `{ ...tex, sampler }`. Both inherit `_recoverySource` without passing through the stamp, so neither wrapper was tracked and both could survive recovery holding the lost device's `GPUTexture`.

A follow-up review then found the sharper case: a clone doesn't keep its base alive. If the app holds a clone and drops the base, there's no live tracked wrapper for that source at all, so nothing initiates the rebuild in the first place.

Fixed in 911f9291 by tracking derived wrappers as captured textures in their own right, rather than in a side registry:

- `cloneTexture2D` notifies a module-level hook the capture module installs. It has to be a hook because neither `cloneTexture2D` nor the glTF `wrapTexture`/`wrapTexCoord` hooks calling it are handed an `engine`, and a scene that never enables recovery shouldn't carry the bookkeeping.
- The capture module records which recovery state captured each source and adds the derived wrapper to that engine's tracked set. The state is held via `WeakRef` so an app keeping a texture past its engine doesn't pin the engine's registrations and the whole scene graph behind them.
- Putting them in the existing tracked set is what makes the base-collected case work, and they get compacted on the same schedule as everything else.

Cost is unchanged: rebuilding is deduped on the shared *source*, so a whole wrapper family is still one upload and a `url` source isn't re-fetched per wrapper. A wrapper reached mid-rebuild awaits it rather than adopting the lost device's texture, and adoption resolves each wrapper's own captured sampler descriptor — so the glTF sampler wrapper keeps its own wrap/filter settings instead of inheriting the base's.

Four unit tests cover it, including the base-collected case. I checked they actually bite: with the hook stubbed out, 3 of the 4 fail (the fourth covers source-level dedupe, which is independent).

One side effect worth flagging: the reworked version is 114 bytes *smaller* than the first attempt, but scene164 still lands at 143.30 KB, so the ceiling ask moved 143 → 144. Still needs @deltakosh's sign-off.

_Auto-resolved by the [GitHub Copilot app](https://github.com/github/app)._

ryantrem:
Thanks — all four addressed in 06d1b6c0.

**Restore the creator-owned reference.** Real bug, and your three kinds were exactly right: `createTexture2D`, `createTexture2DFromPixels` and `createRenderTexture2D` are precisely the creators that `acquireTexture`. A replacement `GPUTexture` starts at ref-count 0, so a rebuilt texture died on the first consumer that bound and then unbound it:

```
acquireTexture(tex)  // material binds   → 0 + 1 = 1
releaseTexture(tex)  // material unbinds → 1 - 1 = 0 → destroy()
```

`rebuildTexture2D` now re-takes that reference for those three kinds only. `createSolidTexture2D` and the glTF `uploadTex` path take none, so recovery takes none for them; `rebuildDynamicTexture2D` already restored its own; adopting wrappers take none, exactly as `cloneTexture2D` takes none at creation, since they share the reference held for the texture they now point at.

Covered by one test per creator-owned kind, each running a full consumer acquire/release cycle against the rebuilt texture.

**Do not rebuild released or disposed textures.** Also real — a released texture was rebuilt, allocating a replacement and taking a creation-time reference nothing would ever release. `releaseTexture` now clears the wrapper's `_recoverySource` when it drops the last reference and destroys the texture, so recovery leaves it alone. That is the cheapest available signal and it costs nothing measurable: scene50 and scene92 are byte-identical with and without the line (16,173 B / 19,060 B).

Covered by a test that captures a texture, releases it, and asserts recovery calls `createTexture` zero times.

**Keep clones linked through recovery, one rebuild, preserved sharing.** This is the design as of 911f9291 (@RaananW's thread): derived wrappers are tracked as captured textures in their own right, and rebuilding is deduped on the shared recovery **source**, so a wrapper family costs one upload and a `url` source is fetched once. I have added the explicit coverage you asked for — a base plus two clones, asserting `createTexture` is called once and that all three end up on the same `texture`, `view` **and** `sampler`, with their distinct UV transforms intact.

One deliberate exception to "adopt the base's sampler": a wrapper carrying its own captured sampler descriptor gets that descriptor rebuilt instead. The glTF sampler path derives a wrapper specifically to give the same image different wrap/filter settings, so handing it the base's sampler would silently change how the asset renders after a recovery. A wrapper that shared the base's sampler still ends up on the identical object.

**Docs.** `50-device-lost-recovery.md` updated: the fail-fast behaviour replaces the old "context kinds without a registered recovery handler are left untouched" contract in both the Purpose and Coordinator Dispatch sections, the up-front rebuild of captured textures is now in the dispatch sequence, and there is a new **Captured textures** subsection covering weak tracking, source-level dedupe, derived wrappers, and the ownership rules above. In the Test Specification I replaced the obsolete skipped-unregistered-kinds requirement with the fail-fast requirement and added the captured-texture requirements.

All four fixes were checked against buggy code, not just written green — reverting the two ownership changes fails all four new ownership tests.

Validation: 235 files / 1917 unit tests, lint + 6/6 typechecks, scene164 at 143.40 KB.

That last number is the one open item: the scene164 ceiling ask is now **143 → 144** and still needs your call. Nothing else in the repo moves.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.192 - merge @ 472d4dd47f213516b1276b9859c925ac61dbdf23</sub>

deltakosh (changes requested):
Invalidate the recovery source for the entire wrapper family when the shared GPU texture is destroyed. A sibling wrapper currently remains tracked and can rebuild a disposed texture. Add a regression test where a base and clone share one texture, the clone performs the final release, and recovery allocates nothing.

ryantrem:
Bundle size CI was right and my earlier "no other ceiling moves" claim was wrong — fixed in 3f9aa22 (also merged current `master`).

I had measured scene50 and scene92 with and without the always-bundled line and got byte-identical results. Those two scenes just never load the affected modules, so that was a bad control. Five other scenes went over by 8–48 bytes.

The cause turned out to be structural rather than a line count. Giving `gltf-sampler-desc` a **runtime** import of `texture-2d` — it previously imported only types — created an import edge that pulls `texture-2d` into bundles that never load it at runtime. scene248 gained a 92-byte `texture-2d` chunk it did not have at all, plus 63 bytes in `gltf-sampler-desc`: 173 bytes against 140 bytes of headroom.

Both new costs moved off the always-bundled path:

- `gltf-sampler-desc` has an engine in scope, so it registers its derived wrapper through the existing `engine._dlr` capture seam and keeps its type-only import. `cloneTexture2D` has no engine — public API, reached from glTF through `GltfFeature.wrapTexture` — so it keeps the module-level hook, now inlined. Both call the same tracking function.
- Recovery reads the owner count through a tree-shakeable `_textureRefCount` instead of `releaseTexture` clearing `_recoverySource`. For the three creator-owned kinds a zero count means the last release already destroyed the texture, which is what the flag stood for. Kinds whose creator never acquires are exempt, since zero is their steady state. This is a strictly better answer to @deltakosh's second ask: same behaviour, and scenes that never recover carry none of it.

| scene | before | after | headroom on master |
|---|---|---|---|
| scene12 | +69 B | **+1 B** | 28 B |
| scene27 | +66 B | **+15 B** | 43 B |
| scene73 | +70 B | **+1 B** | 62 B |
| scene248 | +173 B | **+38 B** | 140 B |
| scene253 | +134 B | **+50 B** | 86 B |

All five are back under their **unchanged** ceilings. The `texture-2d` chunk delta is gone entirely; the remainder is the registration in `gltf-sampler-desc` itself.

One fixture change worth flagging: `pixelsTexture` now takes the ownership reference its real creator takes. Without it the fixture modelled a *released* texture rather than a live one, which is why eight tests correctly failed the moment recovery started reading that count.

2629 unit tests pass, lint and 6/6 typechecks clean, scene164 at 143.5 KB.

The scene164 ceiling ask is unchanged at **141.7 → 144** and still needs your call — it remains the only ceiling this PR touches.

ryantrem:
Regression test added in 6364399 — but the invariant already held, and for a reason worth spelling out, since the diff doesn't look like what you asked for.

There's no per-wrapper flag to invalidate any more. An earlier revision of this PR did clear `_recoverySource` in `releaseTexture`, which is exactly the design your comment breaks — it marks the wrapper that released, not the family, so a sibling survives still eligible for rebuild. That got replaced while fixing a bundle-size regression: recovery now reads the owner count through `_textureRefCount` instead, and that count is keyed on the `GPUTexture`, not the wrapper.

A derived family shares one `GPUTexture` — `cloneTexture2D`'s `transform` parameter is typed to UV fields only, so a clone cannot point somewhere else — so they share one count. The final `releaseTexture` deletes that single entry, and every wrapper in the family then reads zero and is skipped, whichever one recovery happens to visit first and whichever one performed the release. The family-wide behaviour you're asking for falls out of where the count lives.

The new test is the one you specified: base and clone over one texture, clone performs the final release, recovery allocates nothing. It also asserts both wrappers are tracked beforehand, so it proves the skip rather than passing because tracking missed the clone. I confirmed it fails if the released-texture guard is removed.

26 unit tests in that suite, lint and 6/6 typechecks clean. The architecture doc now records that the count is keyed on the texture rather than the wrapper, and why that matters for derived families.

deltakosh (changes requested):
One blocker remains in released-texture recovery. The skip guard only covers `url`, `pixels`, and `render`, coupling final-release detection to the generic re-acquire path. Dynamic textures are creator-owned but omitted, while bitmap/glTF and solid families can also reach zero after their last renderable releases them. Recovery can therefore rebuild and re-own destroyed texture families. The new pixels-only base/clone test does not exercise those failing kinds.

Separate released-family detection from re-acquisition, preserve tree-shaking, and extend the base/clone final-release regression coverage across the affected recoverable kinds.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.202 - merge @ d0f219895f99c3ea90889d7bfff526c91f98db73</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 72 — NME PBR Full (D8AK3Z)<br/>`scene72` | 106 KB | 108 KB | -2 KB |
| Scene 263 — NPE Particles: Emitters - Sphere<br/>`scene263` | 40 KB | 42 KB | -2 KB |
| Scene 264 — NPE Particles: Change - Size<br/>`scene264` | 38 KB | 40 KB | -2 KB |
| Scene 276 — NPE Particles: Animations 2 (sprite sheet)<br/>`scene276` | 42 KB | 44 KB | -2 KB |
| Scene 277 — NPE Particles: Attractor<br/>`scene277` | 40 KB | 42 KB | -2 KB |
| Scene 280 — NPE Particles: Flow Map<br/>`scene280` | 40 KB | 42 KB | -2 KB |
| Scene 283 — NPE Particles: Multiply Blend<br/>`scene283` | 40 KB | 42 KB | -2 KB |
| Scene 301 — NPE Sprite2D Blend Modes<br/>`scene301` | 36 KB | 38 KB | -2 KB |
| Scene 302 — NPE Moving Emitter<br/>`scene302` | 53 KB | 55 KB | -2 KB |
| Scene 1 — BoomBox PBR<br/>`scene1` | 87 KB | 88 KB | -1 KB |
| Scene 8 — HDR Glass Sphere<br/>`scene8` | 72 KB | 73 KB | -1 KB |
| Scene 9 — Sponza (.babylon)<br/>`scene9` | 59 KB | 60 KB | -1 KB |
| Scene 24 — Hill Valley<br/>`scene24` | 58 KB | 59 KB | -1 KB |
| Scene 27 — Material Variants<br/>`scene27` | 82 KB | 83 KB | -1 KB |
| Scene 32 — KHR_materials_unlit<br/>`scene32` | 79 KB | 80 KB | -1 KB |
| Scene 34 — KHR_node_visibility + KHR_animation_pointer<br/>`scene34` | 97 KB | 98 KB | -1 KB |
| Scene 35 — EXT_mesh_gpu_instancing<br/>`scene35` | 83 KB | 84 KB | -1 KB |
| Scene 39 — KHR_animation_pointer (Animated Waterfall)<br/>`scene39` | 109 KB | 110 KB | -1 KB |
| Scene 99 — Bone Control<br/>`scene99` | 91 KB | 92 KB | -1 KB |
| Scene 41 — Physics Shape Debug Viewer<br/>`scene41` | 92 KB | 93 KB | -1 KB |
| Scene 56 — Axis-Locked Billboards<br/>`scene56` | 58 KB | 59 KB | -1 KB |
| Scene 60 — NME Flat Colour<br/>`scene60` | 54 KB | 55 KB | -1 KB |
| Scene 65 — NME Shadow Receive<br/>`scene65` | 73 KB | 74 KB | -1 KB |
| Scene 67 — NME PBR Core<br/>`scene67` | 75 KB | 76 KB | -1 KB |
| Scene 69 — NME PBR Sheen<br/>`scene69` | 89 KB | 90 KB | -1 KB |
| Scene 73 — Wheel PBR vs NME Viewports<br/>`scene73` | 140 KB | 141 KB | -1 KB |
| Scene 78 — NME Scalar/Vector Math<br/>`scene78` | 56 KB | 57 KB | -1 KB |
| Scene 83 — NME Normals<br/>`scene83` | 68 KB | 69 KB | -1 KB |
| Scene 86 — NME Scene State<br/>`scene86` | 58 KB | 59 KB | -1 KB |
| Scene 98 — Billboard Additive Blend<br/>`scene98` | 58 KB | 59 KB | -1 KB |
| Scene 112 — KHR_texture_basisu Flight Helmet<br/>`scene112` | 119 KB | 120 KB | -1 KB |
| Scene 115 — Alien Picking Frame 100<br/>`scene115` | 126 KB | 127 KB | -1 KB |
| Scene 118 — Billboard Sprite Picking<br/>`scene118` | 75 KB | 76 KB | -1 KB |
| Scene 122 — Gaussian Splatting SOG<br/>`scene122` | 47 KB | 48 KB | -1 KB |
| Scene 127 — Gaussian Splatting Depth Rendering<br/>`scene127` | 60 KB | 61 KB | -1 KB |
| Scene 128 — Gaussian Splatting Depth Rendering (Alpha-blended)<br/>`scene128` | 60 KB | 61 KB | -1 KB |
| Scene 140 — NME PCF Alpha Discard Shadows<br/>`scene140` | 90 KB | 91 KB | -1 KB |
| Scene 141 — ESM Material Casters<br/>`scene141` | 118 KB | 119 KB | -1 KB |
| Scene 143 — Pipelined Post-processes<br/>`scene143` | 69 KB | 70 KB | -1 KB |
| Scene 145 — Geometry Renderer Task<br/>`scene145` | 88 KB | 89 KB | -1 KB |
| Scene 146 — PBR Geometry Renderer (Sponza)<br/>`scene146` | 109 KB | 110 KB | -1 KB |
| Scene 149 — Node-Material Geometry Renderer (PowerPlant)<br/>`scene149` | 107 KB | 108 KB | -1 KB |
| Scene 165 - ShaderMaterial Thin Instances<br/>`scene165` | 44 KB | 45 KB | -1 KB |
| Scene 167 — PBR Lightmap<br/>`scene167` | 75 KB | 76 KB | -1 KB |
| Scene 168 — Mirrored Double-Sided Winding<br/>`scene168` | 77 KB | 78 KB | -1 KB |
| Scene 171 - Navigation Crowd Path<br/>`scene171` | 92 KB | 93 KB | -1 KB |
| Scene 206 - LWR Cutout Billboards<br/>`scene206` | 60 KB | 61 KB | -1 KB |
| Scene 214 — Cascaded Shadow Maps<br/>`scene214` | 67 KB | 68 KB | -1 KB |
| Scene 224 — Bounding Box Gizmo<br/>`scene224` | 78 KB | 79 KB | -1 KB |
| Scene 240 — AnimatedTriangle<br/>`scene240` | 90 KB | 91 KB | -1 KB |
| Scene 244 — PotOfCoalsAnimationPointer<br/>`scene244` | 133 KB | 134 KB | -1 KB |
| Scene 245 — RecursiveSkeletons<br/>`scene245` | 101 KB | 102 KB | -1 KB |
| Scene 248 — TextureSettingsTest<br/>`scene248` | 76 KB | 77 KB | -1 KB |
| Scene 250 — VirtualCity<br/>`scene250` | 91 KB | 92 KB | -1 KB |
| Scene 253 — AnimateAllTheThings<br/>`scene253` | 153 KB | 154 KB | -1 KB |
| Scene 256 — NormalTangentTest<br/>`scene256` | 76 KB | 77 KB | -1 KB |
| Scene 258 — Interleaved UV<br/>`scene258` | 82 KB | 83 KB | -1 KB |
| Scene 259 — Emissive Material<br/>`scene259` | 76 KB | 77 KB | -1 KB |
| Scene 262 — NPE Particles: Basic Properties - Size<br/>`scene262` | 39 KB | 40 KB | -1 KB |
| Scene 265 — EnvironmentTest (EXT_lights_image_based)<br/>`scene265` | 79 KB | 80 KB | -1 KB |
| Scene 279 — Updated Instanced Line System<br/>`scene279` | 47 KB | 48 KB | -1 KB |
| Scene 281 — NPE Particles: Noise Texture<br/>`scene281` | 40 KB | 41 KB | -1 KB |
| Scene 282 — Standard Material UV Transform<br/>`scene282` | 46 KB | 47 KB | -1 KB |
| Scene 284 — NPE Particles: MultiplyAdd Blend<br/>`scene284` | 40 KB | 41 KB | -1 KB |
| Scene 300 — NPE on Sprite2D<br/>`scene300` | 33 KB | 34 KB | -1 KB |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.202/lab/index.html)

<sub>Build 20260825.202 - merge @ d0f219895f99c3ea90889d7bfff526c91f98db73</sub>

ryantrem:
@deltakosh Good catch on the guard — you were right that it was answering the wrong question, and chasing it down turned up a second ownership bug next to it. Both are fixed in `d9c4686a` and `97761d0d`.

## `dynamic` was a real bug

The old guard reused the re-acquire predicate (`url`/`pixels`/`render`), and `dynamic` is creator-owned too — `createDynamicTexture` acquires at `dynamic-texture.ts:128`. It was excluded from re-acquire for a different reason (its own rebuild module restores its reference at `dynamic-texture-recovery.ts:78`, so acquiring again would double it), and reusing that predicate silently dragged the exclusion into the release check. A released dynamic texture was rebuilt. That's exactly the "these should be separated" point, and they now are.

## Solid and bitmap: agreed they can reach zero, but the guard couldn't be extended to them as written

`releaseTexture` is public API and its `(m.get(t) ?? 1) - 1` default means the first release on a never-acquired texture destroys it, so solid and glTF-uploaded textures really can end up destroyed. The problem is that with a two-state map they're *indistinguishable* from their normal live state — both read "no entry", because their creators never acquire.

I checked this rather than assuming: extending the old guard to treat "no owner" as released fails 3 tests including both glTF sampler recovery tests. Every glTF-uploaded and solid texture in a scene would be skipped and left on the lost device, which is the original use-after-free at scale.

So the fix is to make the states distinguishable. `releaseTexture` now leaves a `0` entry behind instead of deleting, giving three states:

| map entry | meaning |
|---|---|
| absent | nothing ever took ownership (solid, glTF upload) |
| positive | owned |
| zero | every owner released it, `releaseTexture` destroyed it |

The key is weak either way, so the zero retains nothing. The guard is now kind-independent — `_isTextureReleased(tex)` — and covers all six kinds. Re-acquisition keeps its own separate predicate, which is the separation you asked for.

## A second bug: ownership came back short for clones

While separating those two questions I found recovery was restoring exactly one reference. `cloneTexture2D` documents that acquire/release pairing is the caller's, so a family can legitimately hold several references to one `GPUTexture`. Pre-loss count of 2, post-recovery count of 1 — the application then makes the two releases it correctly owes, and the first one destroys the texture while the clone is still pointing at it.

Verified with a throwaway probe before fixing it:

```
PROBE release #1 destroyed: true
PROBE clone still points at destroyed texture: true
```

Recovery now reads the outgoing texture's owner count before the rebuild replaces the handle and tops the replacement up to it. Counting instead of re-acquiring also deletes the kind list entirely: `dynamic`'s self-restore is counted rather than doubled, and kinds that never acquired have nothing to carry.

## Tests

Released-texture detection is now a 6-kind × 2-scenario table (12 tests) rather than two hand-written cases, plus tests for the clone family's second reference surviving and `dynamic` not being double-acquired. Every new test was mutation-verified:

| mutation | fails |
|---|---|
| restore the old kind-limited guard | 6 — exactly `dynamic`/`solid`/`bitmap` |
| `delete` instead of `set(t, 0)` on destroy | 12 — every skip test |
| treat "never owned" as released | 3 — including both glTF sampler tests |
| creator-only re-acquire | 1 — the clone family test |
| flat re-acquire for every kind | 2 — clone family + dynamic |

39 tests in the suite, 2640 unit tests green. Bundle: no ceiling regressions — the scenes actually gained a couple of bytes, since `m.set(t, 0)` minifies smaller than `m.delete(t)`.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260825.213 - merge @ e0e78091f2bc40160b284c6e5c7ec29b23a73859</sub>

deltakosh (changes requested):
The released-family fix now covers every recoverable kind, but the ownership top-up carries live renderable references and the recovery handler then reacquires those same bindings. After recovery this inflates refcounts, prevents final disposal from destroying textures, and can disable released-texture detection on a later loss. Carry only references that survive recovery, or settle the abandoned renderable references before snapshotting, and add coverage that includes group re-acquisition plus final scene disposal.

bjsplat:
## Bundle Size Changes

### Increases

| Scene 152 — Unified AnimationManager<br/>`scene152` | 92 KB | 91 KB | **+1 KB** |
| Scene 164 - Device Lost Recovery<br/>`scene164` | 142 KB | 141 KB | **+1 KB** |
| Scene 176 - MosquitoInAmber<br/>`scene176` | 104 KB | 103 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.213/lab/index.html)

<sub>Build 20260825.213 - merge @ e0e78091f2bc40160b284c6e5c7ec29b23a73859</sub>

ryantrem:
You're right, and it reproduces cleanly. Probe against `97761d0d` — a url texture held by its creator plus one mesh, a loss, then every release that will ever happen:

```
PROBE destroyed after the two real releases: false false
PROBE gpu destroy calls: 0
PROBE _isTextureReleased: false
```

| step | count |
|---|---|
| pre-loss: creator + renderable | 2 |
| top-up snapshots `owners=2`, restores the replacement to 2 | 2 |
| `rebuildSceneGpu` abandons the old reference and re-acquires | **3** |
| every release that will ever happen | rests at **1** |

It never reaches 0, so it is never destroyed and `_isTextureReleased` stays false — disabling the released-texture detection added earlier in this PR on any later loss. Exactly as you described.

## Fix — `c85dbb51`

`rebuildTexture2D` now queues `[texture, owners]` instead of carrying it, and `runDeviceLostRecovery` drains the queue after every handler in `_recoverOrder` has run:

```ts
try {
    for (const handler of orderedHandlers) {
        await runRecoveryStep(`running "${handler._kind}" recovery`, () => handler._recover(engine));
    }
} finally {
    settleTextureOwnership?.();
}
```

That is your first option — carry only what survives recovery — obtained structurally rather than by tracking. Whatever is still missing once the handlers have run is *by definition* a reference recovery did not re-establish, so there is no provenance to record and no ordering to get right.

It stays a top-up (`for (let held = _textureOwners(tex); held < owners; held++)`) rather than a flat acquire, so a handler that already re-took its own reference is counted rather than doubled — while a derived clone family, which no handler walks, still comes back at 2 instead of 1.

| shape | pre-loss | after rebuild | after handlers | after settle |
|---|---|---|---|---|
| solid + 1 renderable | 1 | 0 | 1 | 1 |
| url + 1 renderable | 2 | 0 | 1 | 2 |
| clone family, no renderable | 2 | 0 | 0 | 2 |
| dynamic | 1 | 0 | 1 | 1 |

## On the second option — settling before the snapshot

I tried it and backed it out: it is strictly worse than the bug it fixes.

A solid or bitmap texture's creator takes no reference, so one bound by a single renderable sits at 1. Draining early takes it to 0 — and `releaseTexture` writes the zero entry that marks a texture destroyed. `_isTextureReleased` then reports true, `rebuildTexture2D` early-returns, and the texture is **not rebuilt at all**. The re-acquired bind group goes on to draw with the dead device's `GPUTexture`. That trades a refcount leak for the use-after-free this PR exists to fix.

## Coverage

`carries only the ownership recovery does not re-establish itself` covers group re-acquisition *and* final disposal, not just the resting count: it acquires as the creator, captures, registers a handler whose `_recover` re-acquires the way `rebuildSceneGpu` does, then asserts the first `releaseTexture` returns `false`, the second returns `true`, `destroy` was called exactly once, and `_isTextureReleased` is true afterwards.

Mutation-tested both directions:

| mutation | failures |
|---|---|
| carry at rebuild time again (the reported bug) | 1 — the new test only |
| flat `acquireTexture` in the settle instead of the top-up | 2 — clone family + dynamic |

40 tests in the file, eslint and 6/6 typechecks clean, and all six near-ceiling scenes still under budget.

One note on the `finally`: the queue is module state that outlives a failed recovery. Stranding it would pin every rebuilt wrapper strongly — it is the only strong reference in an otherwise weak design — and the next recovery would drain it onto a different device's textures.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260826.26 - merge @ d637abd5dcbd7f311ca1c3aad3cf1d48a5b278bf</sub>

deltakosh (changes requested):
One blocker remains in the ownership settlement fix.

`texture-recovery.ts` keeps `_pendingOwnership` in one module-level array shared by every engine and recovery run, and each run receives the same unscoped settlement function. If two engines recover concurrently, their async rebuilds append to the same queue. Whichever run finishes its handlers first drains both engines' entries, so the other engine is topped up before its handlers reacquire and recreates the refcount inflation/leak this change is meant to prevent. A failure in either run can drain the other run's entries too.

Keep pending ownership local to one recovery/engine/device (for example, return a closure over a per-run list or key it by recovery/device), and add an interleaved two-engine regression test proving one run cannot settle another run's textures.

bjsplat:
## Bundle Size Changes

### Increases

| Scene 164 - Device Lost Recovery<br/>`scene164` | 143 KB | 141 KB | **+2 KB** |
| Scene 152 — Unified AnimationManager<br/>`scene152` | 92 KB | 91 KB | **+1 KB** |
| Scene 176 - MosquitoInAmber<br/>`scene176` | 104 KB | 103 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260826.26/lab/index.html)

<sub>Build 20260826.26 - merge @ d637abd5dcbd7f311ca1c3aad3cf1d48a5b278bf</sub>

ryantrem:
You're right, and the hazard is one the code already half-named — the `finally` comment warned that a stranded queue "outlives this recovery" and a *later* recovery would drain it onto a different device's textures. That framing was sequential. Concurrent is the case that actually matters: a lost GPU process loses every device on the page at once, so two engines entering `runDeviceLostRecovery` together is the ordinary path, not an exotic one. Same defect, worse odds. Fixed in eb532087.

**The trace I confirmed before changing anything**

- A's rebuild pushes `[A, 2]`, then `await`s a handler and yields.
- B's rebuild pushes `[B, 2]` into the *same* array. B's handler acquires B, B's `finally` runs and drains **both** entries. A's replacement is at 0 — its handler hasn't run — so it's topped up to 2.
- A's handler then re-acquires: **A = 3**. A's `finally` finds an empty queue and no-ops.
- Every release that will ever happen leaves A at 1. Never destroyed, and `_isTextureReleased` stays false forever — so a later loss won't even skip it. That's the round-9 inflation, moved from within one engine to across two.

The `catch` in `rebuildRecoverableTextures` had the same reach: one engine's rebuild failure settled the other's queue.

**What I did**

The queue now lives on that engine's `DeviceLostRecoveryState` and the returned settle callback closes over it.

Per-engine is also per-run here — `arm()` re-arms only once the run resolves, so an engine can't overlap itself. I checked that rather than assuming it.

I went with the state object over threading a per-run list as a parameter, and that choice is load-bearing. `rebuildTexture2D` has three call sites, and two of them — `recovery-rebuild.ts` and `sprite-recovery.ts` — are the handlers' own walks, running *during* the handler phase. A parameter those two didn't pass would silently drop the ownership of whatever they rebuilt, which is a quieter version of the bug you just caught. Reading it off the engine keeps every caller correct with no signature churn, and matches how `recoverCapturedSampler` already reaches `engine._deviceLostRecovery?._samplerDescriptors`.

I considered the device-keyed `WeakMap` you suggested as the alternative — it does work, since each `requestDevice()` hands back a distinct device — but it costs more bytes and leaves module-level state in place, which is the thing worth removing.

**Regression test**

`settles only its own engine's textures when two recoveries interleave` — two engines, A parked inside its handlers while B runs start to finish, then asserts neither settled the other's textures. The barrier is the handler's own entry, so there's no counting of microtask ticks.

I mutation-tested it with the internal-field assertions *disabled*, so only behaviour was under test: restoring the shared module array fails it on `expect(releaseTexture(textureA)).toBe(true)` — A resting at 3, the real symptom. The coverage isn't just bookkeeping.

**Cost**

Nothing on the always-bundled path: the field is optional and `getState` never initializes it, the new import is `import type` and erases, and `texture-recovery.ts` is a lazy chunk. All six tightest scenes measure byte-identical to the previous commit. 41 unit tests, lint and 6/6 typechecks green.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260826.29 - merge @ 399be6d2123f84353aad84fdc7414099326f0f97</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 164 - Device Lost Recovery<br/>`scene164` | 143 KB | 141 KB | **+2 KB** |
| Scene 152 — Unified AnimationManager<br/>`scene152` | 92 KB | 91 KB | **+1 KB** |
| Scene 176 - MosquitoInAmber<br/>`scene176` | 104 KB | 103 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260826.29/lab/index.html)

<sub>Build 20260826.29 - merge @ 399be6d2123f84353aad84fdc7414099326f0f97</sub>

deltakosh (changes requested):
One blocker remains in the per-engine ownership settlement.

The new `DeviceLostRecoveryState._pendingOwnership` queue is scoped only to an engine and its safety comment relies on a false invariant: one engine can overlap itself. `runDeviceLostRecovery` replaces `engine._device` before its handlers finish, while `_armedDevice` still points to the lost device. If another recovery kind is enabled during that window, `_enableDeviceLostRecovery` calls `arm()`, which arms the replacement device. A loss of that device before the first run settles starts a second run on the same engine. The runs then share and drain the same `_pendingOwnership` queue, so one can top up the other's textures before its handlers reacquire and recreate the ownership inflation/leak this settlement is meant to prevent. The overlapping runs also contend for the engine device and surfaces, so scoping the queue alone is not sufficient.

The overlap is in the pre-existing coordinator, but this PR introduces shared state that explicitly depends on it being impossible and claims that invariant without enforcing it. The new regression test covers two engines calling `runDeviceLostRecovery` directly; it never drives the real `arm()` path or same-engine overlap.

Serialize or guard recovery runs per engine (or defer rearming while one is active), and add a coordinator-driven same-engine interleave regression test.

ryantrem:
You're right, and the part I got wrong is the part I asserted rather than checked.

I reasoned about `arm` as if the post-run continuation were its only caller. It isn't — `_enableDeviceLostRecovery` calls it too, on every registration. So the window is exactly as you describe: `runDeviceLostRecovery` assigns `engine._device = D2` while `_armedDevice` still names `D1`, and any registration enabled in that window finds `_armedDevice !== engine._device`, arms `D2`, and a loss of `D2` starts a second run on the same engine. A crash-looping GPU process produces that back-to-back loss without any exotic setup.

The comment I wrote — "one engine cannot overlap itself: it re-arms only after its run resolves" — described a coincidence of one code path, not an invariant, and then new shared state was built on top of it. That's the failure worth naming, not just the overlap.

And your last sentence is the one that changed the fix: scoping the queue alone was never going to be enough, because the two runs also contend for `engine._device` and `engine.surfaces`. So I took the second remedy you offered rather than the first.

## The fix — defer arming while a run is in flight

`_recovering` on `DeviceLostRecoveryState`; `arm` refuses while it is set. Set before the `_onLost` dispatch, cleared in both continuations:

- **Success** clears, then re-arms. Nothing is dropped by having waited: `GPUDevice.lost` is a promise, so a device lost during the window resolves for that later subscriber too, and the deferred recovery runs immediately afterwards — serialized behind the run that was already in flight rather than alongside it.
- **Failure** clears without re-arming. A failed run can leave the engine on a device that is already lost, and arming it would spin recovery forever. This preserves today's terminal-failure behaviour exactly; the clear only restores the ability of a later `_enableDeviceLostRecovery` to arm, which it had before this change.

The field is optional, so `getState` is untouched and the always-bundled cost is the guard plus three assignments.

I considered keying the queue by device instead, and rejected it: it makes the queue safe while leaving two runs fighting over one engine's device and surface list — the failure mode you called out in your last line.

## Test

You're right that the existing interleave test never touches the coordinator — it calls `runDeviceLostRecovery` directly. The new one goes through `arm`:

1. A `losableDevice()` helper whose `lost` the test resolves.
2. Lose `D1` for real; the sprite registration's handler parks mid-recovery.
3. While parked, enable a `text-renderer` registration and lose `D2`.
4. Assert `requestDevice` has been called **once** and `engine._device` is still `D2` — no second run.
5. Resume, and assert the deferred loss **is** recovered (`requestDevice` twice), so this defers rather than drops.

Mutation-tested: dropping `|| state._recovering` fails it with `expected 1 times, got 2`.

One caveat I'd rather state than let you discover. I also disabled those two assertions to see whether the ref-count assertions caught the overlap on their own. **They didn't** — in this interleaving the two runs' ownership math still balances. So I rewrote that comment: the ref-count block is an end-state sanity check, and the call-count and device-identity assertions are the actual detector. I'd rather the test say what it really proves.

## Validation

| check | result |
|---|---|
| `test:unit device-lost-recovery` | 42 passed (was 41) |
| `test:unit bundle` | 25 passed, 3 skipped |
| `lint` | clean, 6/6 typechecks |
| bundle, six tightest scenes | scene164 +30 B (224 B → **194 B** headroom); other five byte-identical |

`scene164.ts` is the only lab scene that enables device-lost recovery, so it is the only one affected — no ceiling change needed for this round.

Pushed as `3e2d574f`.

Still outstanding from an earlier round: scene164's `maxRawKB: 144` needs your explicit sign-off — it is raised in this branch but I am not treating that as approved.

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/611/)

<sub>Build 20260827.4 - merge @ 7fbaf6f81c4ddc77902d7200b127ac573264d113</sub>

deltakosh (approved):
No blocking findings. The same-engine overlap issue is resolved: recovery now defers arming replacement devices while a run is in flight, rearms after successful settlement, and the new coordinator-driven regression covers a replacement device lost during the first recovery. The broader callback-isolation behavior predates this change and is not regressed here.

bjsplat:
## Bundle Size Changes

### Increases

| Scene 164 - Device Lost Recovery<br/>`scene164` | 143 KB | 141 KB | **+2 KB** |
| Scene 152 — Unified AnimationManager<br/>`scene152` | 92 KB | 91 KB | **+1 KB** |
| Scene 176 - MosquitoInAmber<br/>`scene176` | 104 KB | 103 KB | **+1 KB** |

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260827.4/lab/index.html)

<sub>Build 20260827.4 - merge @ 7fbaf6f81c4ddc77902d7200b127ac573264d113</sub>

PR #615, opened 2026-08-25
fix(test): repair master typecheck broken by #597/#601 semantic conflict

RaananW:
## master pipeline fix

`master` fails `tsc -p tests/lite/tsconfig.json` 

```
tests/lite/unit/clustered-light.test.ts(287,13): error TS2353: Object literal may only specify known properties, and '_toneMappingHelpers' does not exist in type 'PbrComposerDeps'.
tests/lite/unit/pbr-lightmap.test.ts(262,13): error TS2353: ...
```

Reproduced on a pristine `origin/master` worktree at `8ff3eb15`, with no local changes.

## Why

This is a semantic conflict between two PRs that were each green, and merged 93 minutes apart:

| PR | Commit | Time | What it did |
| --- | --- | --- | --- |
| #597 | `d130265d` | 06:26 | Reworked the PBR shader variant cache; **removed** `_toneMappingHelpers` / `_toneMappingCall` from `PbrComposerDeps`. Tone mapping now travels as the optional `_tm?: ToneMapping` dep, read internally as `_tm?.helpersWGSL`. Updated every `createPbrComposer` call site that existed on master at the time. |
| #601 | `a16717a5` | 07:59 | Added clustered spot light support, including two `createPbrComposer` call sites that still pass the **old** properties. Its branch was validated before #597 landed, so its own CI was legitimately green. |

Neither PR is wrong. The combination was simply never type-checked, and excess property checking rejects the literals against the new interface.

## Why it survived for weeks — two independent holes

It is worth being precise, because "nothing validates master" is only half the story.

1. **No post-merge validation.** Nothing type-checks `master` after a merge, so the moment the two PRs combined, there was no run to notice. #531 fixes this.
2. **The type-check steps could not fail.** Later PRs *did* build merge commits containing both changes, so their Lint job did run `tsc -p tests/lite/tsconfig.json` against the broken combination — and still reported green. The steps pipe `tsc` through `sed` to emit `##vso[task.logissue]` annotations, and a pipeline exits with the status of its last command. `sed` always succeeds, so `tsc`'s exit code was discarded:

   | invocation | exit |
   | --- | --- |
   | `tsc -p tests/lite/tsconfig.json --noEmit` | **2** |
   | the same piped through `sed` (what CI ran) | **0** ← the bug |
   | with `set -euo pipefail` | **2** |

   That is fixed in #531 (`f915c65d`), which adds `set -euo pipefail` to all seven affected steps.

So hole 2 is why the error was invisible on every PR since, and hole 1 is why it was never caught at the point of merge. This PR fixes the resulting breakage; #531 fixes both holes.

## The fix

Follow the convention #597 established in the call sites it *did* update (see `tests/lite/unit/pbr-uv-transform.test.ts`): replace the two dead properties with `_tm: undefined`.

Behaviour-preserving: the tests passed `""` for both, and an absent `_tm` leaves `_tm?.helpersWGSL` undefined, which the template already defaults to `""`.

## Validation

- `pnpm lint` exits **0** (it currently exits 2 on master)
- All **46 tests** across the four `createPbrComposer` specs pass (`clustered-light`, `pbr-lightmap`, `pbr-post-compose`, `pbr-uv-transform`)

Per repo guardrails, no all-scene / parity / bundle / perf suites were run.

## Relationship to #531

#531 adds post-merge validation on `master`. It cannot go green until this lands, because its PR build type-checks the merge commit and inherits this breakage — precisely the "innocent PR debugs someone else's regression" problem #531 exists to prevent.

Recommend merging this first.


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/615/)

<sub>Build 20260825.32 - merge @ 4e07eb9d832e3bd0121a05c0114dc083f8cb1e1e</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260825.32/lab/index.html)

<sub>Build 20260825.32 - merge @ 4e07eb9d832e3bd0121a05c0114dc083f8cb1e1e</sub>

deltakosh (approved):

PR #639, opened 2026-08-27
Fix runtime blur updates and bundle tooling correctness

Popov72:
## Summary

- rebuild plain and depth-of-field blur shader GPU state when the kernel changes at runtime, without re-recording the frame graph or recreating render targets
- preserve underscore-prefixed ESM exports during cross-chunk Terser property mangling
- make stale demo-bundle cleanup use exclusive longest-slug ownership across configured demos and support bundles
- align the lab editor and lint TypeScript projects with source path mappings and `noEmit`

## Details

Runtime blur kernel changes previously called `record()`, which disposes and recreates task-owned render targets. In bloom, the merge pass retained a bind group referencing the destroyed blur texture view. Kernel updates now rebuild only the shader module, pipeline, and bindings; the same correction is applied to the depth-of-field blur path.

The bundler now reserves underscore-prefixed names exported by any emitted chunk so namespace reads stay compatible with producer export tables. Demo cleanup now resolves ownership against the complete known slug set, preventing prefix-related demos and support bundles from deleting or orphaning each other's chunks.

The lab type-check config now extends the editor config, keeping source aliases, ambient WebGPU/XR declarations, and `noEmit` in one source of truth.

## Validation

- 7 focused runtime/build-tooling regression tests
- `pnpm lint`
- `pnpm build:lib`
- filtered Scene 1 bundle: 87.4 KB raw / 36.9 KB gzip, unchanged from baseline

### DIFF
--- a/lab/tsconfig.json
+++ b/lab/tsconfig.json
@@ -1,8 +1,14 @@
 {
-  "extends": "../tsconfig.base.json",
-  "compilerOptions": {
-    "outDir": "./dist",
-    "rootDir": "."
-  },
-  "include": ["lite/src", "gl/src"]
+    // The project an editor picks up for lab/ and the one `pnpm lint` checks through
+    // tsconfig.typecheck.json, which extends this file.
+    "extends": "../tsconfig.base.json",
+    "compilerOptions": {
+        "noEmit": true,
+        "paths": {
+            "babylon-lite": ["../packages/babylon-lite/src/index.ts"],
+            "babylon-lite/*": ["../packages/babylon-lite/src/*"],
+            "babylon-lite-gl": ["../packages/babylon-lite-gl/src/index.ts"]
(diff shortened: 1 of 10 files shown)


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/639/)

<sub>Build 20260827.10 - merge @ e07cce0dcb7429532917053e2de4aca8f643b5cc</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260827.10/lab/index.html)

<sub>Build 20260827.10 - merge @ e07cce0dcb7429532917053e2de4aca8f643b5cc</sub>

RaananW (approved):

PR #651, opened 2026-08-31
fix(gizmo): keep rotation sector direction across faces

### LINKED ISSUE #636: Rotation gizmo - Pie slice sector flip reversed
### Version
`@babylonjs/lite` **1.22.0**

### What’s broken
When you drag a rotation ring, the yellow “pie slice” (the sector that shows how far you rotated) sometimes goes the **opposite** way from your mouse.

It depends on where the camera is:
- Look at a ring from one side → sector looks correct
- Orbit around so you’re looking at the **back** of that same ring → sector looks mirrored

The object itself can still rotate the right way. It’s the **visual feedback** that’s wrong, which makes it feel like rotation is broken / random.

### Why (simple version)
The sector shader still has this:

```wgsl
let delta:f32 = select(-yAngle, yAngle, frontFacing);
```

That means: “if the camera is looking at the back of the plane, flip the sweep.”

Full Babylon.js does the same kind of flip in the **shader**, but it also flips the angle again in **JavaScript** when the camera is on the other side of the ring (`cameraFlipped` in `planeRotationGizmo.js`). The two flips cancel out, so the sector still matches the mouse.

Lite kept the shader flip, but **dropped** the matching JS flip. So only one flip happens → from the back side, the yellow sector goes the wrong way.

### Expected
The yellow sector should always grow in the same direction as the pointer drag, no matter which side of the ring you’re looking from.

### Actual
From the back side of a ring, the sector sweeps the opposite way from the drag.

### Suggested fix
A sector painted on the ring plane doesn’t need a view-dependent flip at all (like a sticker on glass — it already mirrors when you look from behind).

In `rotation-sector-material.js`, replace:

```wgsl
select(-yAngle, yAngle, frontFacing)
```

with a constant sign (for Lite’s current plane UVs, `u→+x` / `v→+y`, that constant is `-yAngle`):

```wgsl
-yAngle
```

Alternatively, restore the upstream `cameraFlipped` negation on the CPU so it matches the shader again — but the constant-sign approach is simpler and matches how a plane-painted sector should behave.

### How to reproduce
1. Create a rotation gizmo on any mesh
2. Drag a ring while looking at it from the “front” → sector follows the mouse
3. Orbit the camera to the other side of that same ring
4. Drag again → sector goes the opposite way from the mouse

### Playground
https://liteplayground.babylonjs.com/snippet/G6TXNM/v/0

deltakosh:
## Summary
- remove the redundant `front_facing` sign inversion from the rotation-sector shader
- preserve the sector sweep direction when viewing the same ring from either side
- add a focused regression test that pins the view-independent shader contract

## Reproduction
On the Z rotation ring, the same 48-degree counter-clockwise screen gesture produced a clockwise/mirrored sector from the negative-normal camera side, while object rotation remained correct. The opposite camera side rendered the sector correctly.

## Validation
- `corepack pnpm exec vitest run --project unit tests/lite/unit/rotation-sector-material.test.ts`
- targeted ESLint for the changed source and test
- TypeScript checks for the package and Lite tests

Fixes #636

### DIFF
--- a/packages/babylon-lite/src/gizmo/rotation-sector-material.ts
+++ b/packages/babylon-lite/src/gizmo/rotation-sector-material.ts
@@ -19,19 +19,18 @@ const VERTEX_SOURCE = `struct VertexOutput{@builtin(position) position:vec4<f32>
 // Port of BJS rotation sector GLSL:
 //   uv = vUV - 0.5
 //   angle = atan2(uv.y, uv.x) + π
-//   delta = frontFacing ? angles.y : -angles.y
+//   delta = -angles.y
 //   begin = angles.x - delta * angles.z
 //   start = min(begin, begin + delta)
 //   end   = max(begin, begin + delta)
 //   ... wrap and accumulate intensity over 5 periods
 //   colour = vec4(rotationColor, min(intensity * 0.25, 0.8)) * (1 - step(0.5, len))
 const FRAGMENT_SOURCE = `struct VertexOutput{@builtin(position) position:vec4<f32>,@location(0) uv:vec2<f32>,};
-@fragment fn mainFragment(input:VertexOutput,@builtin(front_facing) frontFacing:bool)->@location(0) vec4<f32>{
+@fragment fn mainFragment(input:VertexOutput)->@location(0) vec4<f32>{
 let TWO_PI:f32=6.283185307;
 let uv:vec2<f32>=input.uv-vec2<f32>(0.5,0.5);
 var angle:f32=atan2(uv.y,uv.x)+3.141592;
-let yAngle:f32=shaderUniforms.angles.y;
-let delta:f32=select(-yAngle,yAngle,frontFacing);
+let delta:f32=-shaderUniforms.angles.y;
 let begin:f32=shaderUniforms.angles.x-delta*shaderUniforms.angles.z;
 var startA:f32=select(begin+delta,begin,begin<begin+delta);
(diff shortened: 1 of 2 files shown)


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/651/)

<sub>Build 20260831.25 - merge @ 07374e4346c9aa94f4fc2a8fb82f3ecfe6cc3b7e</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260831.25/lab/index.html)

<sub>Build 20260831.25 - merge @ 07374e4346c9aa94f4fc2a8fb82f3ecfe6cc3b7e</sub>

RaananW (approved):
> 🤖 This review comment was created by an AI agent.

**Verdict: ✅ Approve**
**CI: ⏳ Pending** — all reported build/API, bundle-size, lint/type-check, parity, unit, CodeQL, and secret-scanning checks pass; only `PR Validation Performance: Lite vs Stable` is still running.
**CI safety: ✅ No credential-exposure risk found**

No findings. The change removes the redundant face-dependent shader inversion, preserves the CPU-side signed-angle flow, and adds a focused regression assertion. Existing mid-drag gizmo parity coverage validates the chosen constant sign; the new test locks the view-independent shader contract.

PR #747, opened 2026-09-18
fix(geometry): emit valid normal outputs for unlit Standard materials

sdoleg2011:
## Problem

A `StandardMaterial` with `disableLighting = true` compiles a fragment that never declares the lit-only `normalW` local (`standard-template.ts` emits `var normalW = normalize(input.vn);` only when lighting is enabled).

`composeStandardGeometryShader` referenced `normalW` for `VIEW_NORMAL` and `WORLD_NORMAL` regardless, so as soon as a `GeometryRendererTask` with a normal attachment met one unlit mesh, the geometry pipeline got invalid WGSL:

```
Error while parsing WGSL: :6:52 error: unresolved value 'normalW'
[Invalid ShaderModule] is invalid due to a previous error.
[Invalid RenderPipeline] is invalid due to a previous error.
[Invalid CommandBuffer from CommandEncoder "frame"] is invalid due to a previous error.
```

The last line is what makes it painful: the whole frame's command buffer is rejected, so the scene goes black instead of just losing that one mesh.

I hit this in a kitchen configurator running on `@babylonjs/lite` 1.30.0 through `lite-compat`: LED strips there are unlit emissive boxes (`disableLighting = true`), and adding a geometry pass with `VIEW_DEPTH` + `VIEW_NORMAL` for a screen-space effect blacked the canvas out the moment the first strip appeared. Still reproduces on current `master`.

## Fix

For unlit features the normal attachments read the interpolated vertex normal (`normalize(input.vn)`, always written by the Standard vertex stage). Lit fragments keep `normalW`, which carries the bump perturbation, so their output is unchanged.

## Tests

`tests/lite/unit/shader-integration.test.ts` — new case composes the geometry shader with `DISABLE_LIGHTING` and `VIEW_NORMAL` + `WORLD_NORMAL`; it fails on `master` (the fragment contains `normalW` without a declaration) and passes with the fix. The lit path is asserted to be unchanged.

Ran focused: `shader-integration`, `geometry-renderer-task`, `standard-geometry-features` (81 tests), plus eslint, prettier and `tsc` for `packages/babylon-lite`.

## Bundle

Local per-module accounting for `scene145`: `standard-geometry-output-shader.js` 4106 → 4147 B (+41 B). No ceiling changes.


deltakosh (approved):
Hi Denis, the unlit geometry path now emits valid normal outputs while preserving the lit/bump-normal behavior, and the focused shader regression is clear. Looks good, thanks!

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/747/)

<sub>Build 20260918.37 - merge @ a89061aeb74155a11de111492a411765e1f6488d</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260918.37/lab/index.html)

<sub>Build 20260918.37 - merge @ a89061aeb74155a11de111492a411765e1f6488d</sub>

PR #748, opened 2026-09-18
fix(geometry): load a material family bridge that first appears after preload

sdoleg2011:
## Problem

`GeometryRendererTask._preload` imports the Standard / PBR / Node geometry bridges only for the material families present in the scene at that moment. That is fine for a static scene, but in an editor the scene keeps changing: if the task was preloaded while the scene held only Standard materials and a PBR mesh is added later, `rebuildBoundMeshes` reaches the missing factory and throws:

```
TypeError: task._computePbrFeatures is not a function
```

`_boundVer` is stamped only after a successful rebuild, so the task throws again on every `execute()` and on every `frameGraph.build()` (resize, material group rebuild). In practice the scene stops rendering from the moment the first mesh of the new family appears.

I ran into this in a kitchen configurator on `@babylonjs/lite` 1.30.0: the room starts with Standard materials only and catalogue modules bring PBR and Node materials in later, while a geometry pass is already feeding a screen-space effect. Reproduces on current `master`.

## Fix

In `resolveSourceMaterial`, a mesh whose family bridge is not loaded yet is skipped (the documented behaviour for an unsupported family), the bridge is imported through the existing `_preload` path, and `_boundVer` is reset so the next `execute()` binds the skipped meshes. Families that are already loaded take the same path as before. Meshes that are already bound keep rendering while the import is in flight.

## Tests

`tests/lite/unit/geometry-renderer-task.test.ts` — new case: a Standard-only task, then a PBR mesh is added and the scene version bumped. On `master` `execute()` throws the `TypeError` above. With the fix it does not throw, the Standard mesh stays bound, and once the import settles the PBR bridge is installed and the bound list is marked stale.

Ran focused: `geometry-renderer-task` (26 tests), eslint, prettier, `tsc` for `packages/babylon-lite`.

## Bundle

Local per-module accounting for `scene145`: `geometry-renderer-task.js` 7456 → 7574 B (+118 B). No ceiling changes.


bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/748/)

<sub>Build 20260918.36 - merge @ 1a4c6f9b276f28ee2ac974b9ad796cbf8ba8b3f6</sub>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260918.36/lab/index.html)

<sub>Build 20260918.36 - merge @ 1a4c6f9b276f28ee2ac974b9ad796cbf8ba8b3f6</sub>

packages/babylon-lite/src/frame-graph/geometry-renderer-task.ts
```diff
@@ -847,21 +849,32 @@ function resolveSourceMaterial(task: GeometryRendererTaskInternal, material: Mat
     if (!buildGroup) {
         return null;
     }
-    if (buildGroup._materialFamily === "standard") {
+    const family = buildGroup._materialFamily;
+    // `_preload` imports a family bridge only for the families present when it runs. When a family first
+    // appears later (a PBR mesh added to a Standard-only scene), skip its meshes while the bridge is
+    // imported, then reset `_boundVer` so the next `execute()` binds them.
+    if (!(family === "standard" ? task._createStandardGeometryView : family === "pbr" ? task._createPbrGeometryView : family !== "node" || task._createNodeGeometryView)) {
```
deltakosh, line 876: This guard treats `_createStandardGeometryView` as proof that the whole Standard family is ready, but `_preload()` assigns that factory before awaiting `preloadStandardGeometryFeatures()`. If the first late Standard mesh needs skeletal velocity or thin-instance helpers, a concurrent resize/scene mutation can pass this guard and then throw `... feature was not preloaded`, stopping the render loop even though the import later completes. Please expose family readiness only after all of that family's feature preload has completed successfully.

packages/babylon-lite/src/frame-graph/geometry-renderer-task.ts
```diff
@@ -847,21 +849,32 @@ function resolveSourceMaterial(task: GeometryRendererTaskInternal, material: Mat
...
+        task._lateLoad ??= task._preload!().then(() => {
```
deltakosh: This cached fire-and-forget promise has no rejection path. If a bridge chunk/preload rejects, the rejection is unhandled and `_lateLoad` stays permanently rejected, so `??=` prevents this family (and later families) from ever retrying even after availability recovers. Please handle/report the failure, clear the cached promise on rejection, and leave the task invalidated so a later frame can retry.

deltakosh (changes requested):
Hi Denis, the late-family recovery direction is good, but two asynchronous readiness cases can still stop geometry rendering: Standard is exposed before its optional helpers finish loading, and a rejected late load can never retry. Please tighten those lifecycle paths. Thanks for the focused fix!

sdoleg2011:
Thanks, both points were real — addressed in 637422c7 (branch rebased onto current master).

**Family readiness.** `_preload()` now awaits `preloadStandardGeometryFeatures()` *before* it installs `_createStandardGeometryView` / `_computeStandardFeatures`, so the guard can no longer see a half-ready Standard family. New test holds the feature preload open (the real one runs everywhere else), drives an `execute()` plus a second scene mutation through that window, and checks that the family stays skipped without throwing and becomes available only once the preload resolves. It fails on the previous revision.

**Rejected late load.** The cached promise now has a rejection path: the error is reported through `console.error` (same convention as `shadow-inputs.ts` / `scene-material-swap.ts`), and success and failure end the same way — `_lateLoad` cleared and `_boundVer` reset — so the next frame starts a fresh attempt instead of locking this family and any later one out. New test rejects the first `_preload()`, checks the report, the cleared promise and the stale bound list, then lets the second attempt succeed and install the bridge. It also fails on the previous revision.

One thing I left alone on purpose: a *feature* that first appears for a family that is already loaded (say, the first thin-instanced Standard mesh arriving after the Standard bridge is in) still relies on the original `_preload()` having seen such a mesh. That behaviour is unchanged from master and felt like a separate change; happy to follow up if you want it covered here.

Ran focused: `geometry-renderer-task` (28 tests), `standard-geometry-features` + `shader-integration` (56 tests), eslint, prettier, `tsc` for `packages/babylon-lite`. Local per-module accounting for `scene145`: `geometry-renderer-task.js` +27 B over the previous revision of this PR.

deltakosh (changes requested):
Please keep a late PBR mesh deferred until its runtime PBR group build has published the scene geometry context. In a running Standard-only scene, adding the first PBR mesh queues its asynchronous forward build. If a resize/re-record or unrelated `_renderableVersion` change triggers a geometry rebuild while that forward build is pending, the geometry task starts its separate `_preload()` concurrently. If the bridge import wins, `_createPbrGeometryView` becomes available and the completion handler invalidates `_boundVer`; the next `execute()` then reaches `buildPbrGeometryRenderable()` before `_pbrGeomContext` exists and throws synchronously. That exception is outside the new late-load rejection handler and can stop the render loop rather than retrying once the forward build completes.

The new tests only inspect the imported factory/Promise state, so please add coverage that executes the geometry task while the forward PBR build is still pending and verifies the mesh is included after that build completes.

Merci pour la correction des deux blockers précédents — on y est presque !

sdoleg2011:
Thanks, David — you are right, and that is in fact the very exception I hit while writing the first test and then worked around instead of questioning — `buildPbrGeometryRenderable: scene has no PBR context`.

Fixed in 981d80eb: `rebuildBoundMeshes()` now keeps a PBR mesh out of the bound list until the scene has published a PBR geometry context for it (`_pbrMeshGeomContexts.has(mesh)` or the scene-level `_pbrGeomContext`). No extra scheduling is needed for the wake-up: `materializeRuntimeMesh()` publishes the context and bumps `_renderableVersion` when the forward build completes, so the next `execute()` re-syncs the list and binds the mesh. Standard and Node geometry renderables do not depend on a forward-build context, so the check is PBR-only.

Coverage, as requested, executes the task through the pending window: Standard-only task, late PBR mesh, bridge import completes while the scene still has no PBR context; `execute()` (twice, with an unrelated version bump in between) must not throw and must keep drawing the Standard mesh only. Then the forward build "completes" (context published + version bump) and the next `execute()` builds and binds the late mesh. On the previous revision the test fails with exactly the exception you described. One caveat I want to be upfront about: the final step uses a recording renderable builder on the real PBR geometry view, because the real PBR renderable needs a composed scene shader that the mocked device in this suite cannot provide; the pending-window part runs the real code path end to end.

Ran focused: `geometry-renderer-task` (29 tests), `standard-geometry-features` + `shader-integration` (56 tests), eslint, prettier, `tsc` for `packages/babylon-lite`. Local per-module accounting: `geometry-renderer-task.js` +40 B over the previous revision on `scene145` and `scene146`. Branch rebased onto current master.

deltakosh (changes requested):
The new guard fixes the no-context case, but it still accepts any existing `_pbrGeomContext` before the pending forward rebuild has made that context capable of rendering the late mesh. In an existing PBR scene, add the first thin-instanced PBR mesh and resize/re-record while its asynchronous forward build is pending: geometry binds it against the old global composer, which has no thin-instance helpers, so all instances draw at the base transform. If the late mesh also introduces morph targets, the early geometry shader references undeclared `morphedPos`/`morphedNorm` and the pipeline is invalid.

Please defer the mesh until its applicable forward build has completed rather than treating any global PBR context as sufficient. The new regression starts with no global context and stubs the real geometry builder, so it misses the insufficient-existing-context case.

Merci pour la correction du premier scénario — cette garde doit juste attendre la capacité requise, pas seulement un contexte quelconque.

sdoleg2011:
Thanks, David — understood: the question is not whether *a* PBR context exists but whether the forward build that covers *this* mesh has finished. Reworked in 307c2cd9 (rebased onto current master).

**Guard.** A PBR mesh of the scene is now bound only once its group's tracked forward output (`scene._groups.get(buildGroup).o`) holds a renderable for it. That is exactly what a completed forward build leaves behind in all three paths — the initial build (`scene-core`), the whole-group PBR rebuild a runtime add triggers (`rebuildSceneGroups` sets `o` to the rebuilt renderables), and the per-mesh runtime build (`materializeRuntimeMesh`) — and it is absent while the build is pending: `moveRuntimeMeshToGroup()` puts the mesh into the group list immediately, but no renderable exists for it yet. All of them bump `_renderableVersion` when they finish (the group rebuild also rebuilds the frame graph), so the bound list re-syncs and picks the mesh up without any extra wake-up. The per-group lookup is a `Set` built lazily once per re-sync, so this stays linear on large scenes. The earlier context-existence check is gone, this one subsumes it.

Off-scene meshes of an explicit `config.meshes` list are never forward-built, so they are exempt and keep using the scene-level context as before.

**Coverage.** The pending-window part of the tests runs the *real* PBR geometry builder: with no context a premature bind throws out of `execute()`, and with an insufficient one it shows up as the late mesh appearing in `_bound` too early.

- no PBR context yet (first PBR mesh in a Standard-only scene) — the case from the last round;
- the one you described: an existing, insufficient `_pbrGeomContext`, a late thin-instanced PBR mesh that has joined its group but has no forward renderable yet, `execute()` plus an unrelated version bump in the window → the mesh stays out of `_bound`, the Standard mesh keeps drawing. **This one fails on the previous revision**: the mesh is bound against the old context;
- an off-scene PBR mesh of an explicit list is still bound.

In the first two the forward build then "completes" the way the real one does (context published, renderable tracked in the group's output, version bump) and the next `execute()` binds the late mesh. As before, only that last step uses a recording builder, because the real PBR renderable needs the composed scene shader that the mocked device cannot produce.

Ran focused: `geometry-renderer-task` (40 tests), `standard-geometry-features` + `shader-integration` (56 tests), eslint, prettier, `tsc` for `packages/babylon-lite`. Local per-module accounting for `scene145`: `geometry-renderer-task.js` +123 B over the previous revision.

deltakosh (changes requested):
Hi! The new PBR readiness check still accepts stale make-before-break output. I reproduced an already-built PBR mesh gaining thin instances and undergoing a same-group material rebuild: while that rebuild remained pending, late geometry-bridge completion caused the geometry pass to bind the old context and issue the instanced draw without the instance-matrix buffer. It corrected itself only after the forward rebuild completed, so the intermediate geometry-buffer contents are wrong. Please track readiness for the mesh's current material/capability generation rather than historical `group.o` membership, and cover the retained-old-output window with a regression test. Thanks for continuing to tighten this up!

sdoleg2011:
Thanks, David — agreed: membership in `group.o` only says the mesh was built *at some point*, and with make-before-break it keeps saying so right through a pending rebuild. Reworked in 87a5bbe0 (rebased onto current master) to track the generation instead.

**Generation stamp.** The PBR group builder now stamps every forward renderable with what it was built for: `_gen: [renderFeatures, meshFeatures]` — the material's render-feature object and the mesh capability bits it composed with. Both halves already behave like generation tokens, so nothing new has to be maintained at the request sites:

- `rebuildMaterial()` drops `source._renderFeatures` the moment a rebuild is *requested*, and the next build (forward, or the geometry task's own `resolveSourceMaterial`) creates a fresh object. A pending material rebuild therefore shows up as `stamp !== mesh.material._renderFeatures`; when the forward rebuild completes it stamps its new renderable with that fresh object and the two agree again.
- a capability change (first thin instances, morph targets, skeleton, vertex colours…) shows up by recomputing `_computeMeshFeatures(mesh)` and comparing with the stamped bits (ignoring the pass-dependent receive-shadows bit).

**Readiness.** `isPbrForwardBuildCurrent(forward, mesh)` lives in the lazily loaded `pbr-geometry-view` module and is handed to the task with the PBR bridge, like the view factory. `rebuildBoundMeshes()` binds a PBR mesh of the scene only while the renderable its group tracks for it carries the mesh's current generation — so not before its first forward build, and not in the retained-old-output window. When the forward build completes it bumps `_renderableVersion` and the list re-syncs. Off-scene meshes of an explicit list stay exempt.

One behaviour to be explicit about: in that window a mesh that *was* bound is dropped from the pass until its forward rebuild lands, rather than kept on its previous geometry binding. Carrying a single binding across `rebuildBoundMeshes()` would cut across the make-before-break retirement of bindings and shared views there, and I did not want to widen this PR further; if you would rather retain it, I can look at that separately.

**Coverage.**
- `pbr-forward-generation.test.ts` (new) runs the **real** forward PBR group build on the mocked device: the renderable is stamped with the material's render-feature object; it is current right after the build; it turns stale when the mesh gains thin instances, and when the render-feature object is dropped as `rebuildMaterial()` does; a rebuilt renderable is current again while the retained one stays stale.
- `geometry-renderer-task.test.ts`: the retained-old-output window you reproduced — an already-built PBR mesh (tracked, stamped renderable; existing context) gains thin instances and gets a material rebuild request, then the geometry bridge finishes loading late. The mesh must stay out of `_bound`; each change alone keeps it out; once the group tracks a renderable stamped for the current generation it is built and bound. It fails on the previous revision (the mesh is bound in the window). The two late-mesh cases from the earlier rounds now complete the forward build with a stamped renderable.

Ran focused: 8 files / 139 tests around the forward PBR build and the geometry task (`geometry-renderer-task`, `pbr-forward-generation`, `alpha-to-coverage`, `pbr-shader-variant-cache`, `material-plugin-identity`, `pbr-shadow-caster-receiver`, `standard-geometry-features`, `shader-integration`), eslint, prettier, `tsc` for `packages/babylon-lite`.

Local per-module accounting over the previous revision: +9 B in `pbr-renderable.js` on a plain PBR scene (`scene1`) for the stamp; on the geometry-renderer scenes (`scene145` / `scene146`) +178 / +186 B in total, almost all of it in the lazy `pbr-geometry-view` chunk.

deltakosh (changes requested):
The new generation check still accepts retained output across light/shadow changes. I reproduced enabling `receiveShadows` on a single-light PBR mesh while its rebuild remained pending: the masked comparison passed, but geometry requested the multi-light path from the old single-light composer and emitted invalid WGSL with undeclared light symbols. Adding a second light reproduced the same failure without changing any mesh capability bits. Please include shadow-receiving and scene PBR-context generation in readiness tracking, and add focused regressions for both retained-output windows. Thanks for continuing to work through these async edge cases!

sdoleg2011:
Thanks, David — both windows reproduce here, and the cause is the one you describe: the stamp covered the material and the capability bits only, and the comparison masked the receive-shadows bit, so nothing in it could see a light or shadow change.

Changes in 7af58e2e:

- **The forward stamp is now the whole request the build made of its PBR context**: the context object itself, the mesh feature bits including receive-shadows, the light mode, the single-light type, and the material render-feature object. The context object is created before the renderables so each one can carry it; it is still published at the same point as before.
- **One derivation on the geometry side.** `_pbrMeshRequest(scene, mesh, noShadows)` returns the context the pass would compose against (the per-mesh runtime context, else the scene one) plus mesh features, light mode and single-light type, all from the live scene. `buildPbrGeometryRenderable` now builds from it, and `isPbrForwardBuildCurrent` compares the same call with the stamp, so the predicate and the builder cannot drift apart. The predicate evaluates it under the forward shadow rule (shadow-output materials never receive). The override-camera floating-origin case stays safe, because the forward scan always loads the single-light block of a one-light mesh.
- **Scene PBR-context generation** is covered by context identity: a tracked renderable that was composed against another context than the one the pass is about to use is stale, even when material, capabilities and lights all match. Once a group rebuild commits, the per-mesh contexts are reset and the new renderables carry the new scene context, so the mesh is picked up again on that version bump.

The invariant: the geometry pass only asks of a context what the forward build of that same mesh asked of it.

Regressions:

- `pbr-forward-generation.test.ts` (real forward build on a mocked device): `receiveShadows` enabled on a single-light mesh under a shadow-casting light; a second light added; the single light replaced by a light of another type; output built against another context, including the per-mesh runtime context case.
- `geometry-renderer-task.test.ts`: both retained-output windows at task level. The mesh is bound, the change lands, the re-sync keeps it out without throwing, and it is bound again when the forward rebuild completes. The existing generation test also gained the other-context step.
- With the predicate reduced to the previous comparison, the three new forward-generation cases and three task-level cases fail, and everything else passes.

Two behaviours worth knowing:

- Adding a light does not schedule a forward rebuild by itself. Until the application calls `rebuildScenePbrPipelines`, the affected meshes stay out of the geometry pass instead of being composed against a context that cannot serve them. The forward pass is equally stale in that state, since it keeps its single-light shader.
- From reading `recovery-rebuild.ts`: device recovery rebuilds the groups without resetting `_pbrMeshGeomContexts`, so a runtime-added PBR mesh keeps pointing at its pre-loss context. With this change such a mesh is held out of the geometry pass rather than composed against that context. I left recovery untouched here. I can reset the per-mesh contexts there in this PR or in a follow-up, whichever you prefer.

Bytes against the previous revision: scene1 +12 B, scene145 +87 B, scene146 +99 B, no ceiling changes. Focused vitest (the 12 files touching these modules), eslint, prettier and `tsc` for the package and for `tests/lite` are clean.

deltakosh (approved):
The forward-readiness stamp now covers the PBR context, current material and mesh capabilities, and the light/shadow request derived by the geometry path itself. The retained-output regressions cover the previously failing rebuild windows. This resolves my blocker. Thanks!

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/748/)

<sub>Build 20260919.8 - merge @ d30e89ae34150bec211d9f6ce086b540eb314294</sub>

bjsplat:
## Bundle Size Changes

### Increases

| Scene 147 — Circle of Confusion<br/>`scene147` | 102 KB | 101 KB | **+1 KB** |
| Scene 149 — Node-Material Geometry Renderer (PowerPlant)<br/>`scene149` | 95 KB | 94 KB | **+1 KB** |
| Scene 175 - Navigation Raycast<br/>`scene175` | 91 KB | 90 KB | **+1 KB** |
| Scene 310 — Mesh Blending Core<br/>`scene310` | 132 KB | 131 KB | **+1 KB** |
| Scene 313 — Mesh Blending Alpha Cutouts<br/>`scene313` | 133 KB | 132 KB | **+1 KB** |

### Ceiling headroom

⚠️ **1 scene this PR added or grew now sits under 1.0 KB of headroom.**

| Scene | Size | Ceiling | Headroom | Δ this PR |
|-------|------|---------|----------|-----------|
| Scene 112 — KHR_texture_basisu Flight Helmet<br/>`scene112` | 119.00 KB | 119.90 KB | **926 B** | +20 B |

<details>
<summary>Tightest scenes repo-wide — 1 of 102 under 1.0 KB</summary>

| Scene 112 — KHR_texture_basisu Flight Helmet<br/>`scene112` ⬅ moved by this PR | 119.00 KB | 119.90 KB | 926 B |
| Scene 35 — EXT_mesh_gpu_instancing<br/>`scene35` ⬅ moved by this PR | 82.55 KB | 83.70 KB | 1.1 KB |
| Scene 26 — PBR Subsurface<br/>`scene26` ⬅ moved by this PR | 88.36 KB | 89.60 KB | 1.2 KB |
| Scene 247 — TeapotsGalore<br/>`scene247` ⬅ moved by this PR | 84.64 KB | 85.90 KB | 1.3 KB |
| Scene 248 — TextureSettingsTest<br/>`scene248` ⬅ moved by this PR | 75.42 KB | 76.70 KB | 1.3 KB |
| Scene 31 — KHR_materials_emissive_strength<br/>`scene31` ⬅ moved by this PR | 78.68 KB | 80.00 KB | 1.3 KB |
| Scene 28 — Clearcoat glTF<br/>`scene28` ⬅ moved by this PR | 85.07 KB | 86.40 KB | 1.3 KB |
| Scene 257 — Negative Node Scale<br/>`scene257` ⬅ moved by this PR | 78.85 KB | 80.20 KB | 1.3 KB |
| Scene 259 — Emissive Material<br/>`scene259` ⬅ moved by this PR | 75.55 KB | 76.90 KB | 1.3 KB |
| Scene 32 — KHR_materials_unlit<br/>`scene32` ⬅ moved by this PR | 78.35 KB | 79.70 KB | 1.4 KB |

</details>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260919.8/lab/index.html)

<sub>Build 20260919.8 - merge @ d30e89ae34150bec211d9f6ce086b540eb314294</sub>

PR #750, opened 2026-09-18
fix(compat): shade an unlit StandardMaterial by its emissive colour

sdoleg2011:
## Problem

Babylon.js shades a `StandardMaterial` with `disableLighting = true` as `emissive + ambient`. In `default.fragment` the diffuse colour only scales the accumulated light (`diffuseBase * diffuseColor + emissiveColor + vAmbientColor`), and with lighting disabled `diffuseBase` stays zero.

Lite's unlit path is `emissive × diffuse` (`standard-template.ts`, the `_disableLighting` branch). That matches Babylon.js only while the diffuse colour is white. The very common "unlit colour" setup renders **black** through the compat layer:

```ts
const mat = new StandardMaterial("led", scene);
mat.diffuseColor = Color3.Black();
mat.specularColor = Color3.Black();
mat.emissiveColor = new Color3(1, 0.85, 0.63);
mat.disableLighting = true;
```

In my case these are the LED strips of a kitchen configurator: warm white in Babylon.js, solid black on `@babylonjs/lite-compat` 1.30.0-preview.

## Fix

The compat `StandardMaterial` hands Lite a neutral white diffuse while lighting is disabled, and keeps the app's own diffuse colour for the Babylon.js-facing getter, for `clone()`, and for when lighting is switched back on. Setter order does not matter.

I kept the change in the compat layer on purpose: Lite's own unlit formula is relied on by native code (the bounding-box / camera gizmos use `diffuse = colour, emissive = white`, the physics viewer sets both), so changing the core formula would alter those. If you would rather align the core formula with Babylon.js instead, I am happy to rework this.

Not covered: the `ambientColor × scene.ambientColor` term of the Babylon.js unlit formula. It is zero by default and I did not want to widen the change.

## Tests

`packages/babylon-lite-compat/tests/materials.test.ts` — new case for the setup above: fails on `master` (`_lite.diffuseColor` is `[0, 0, 0]`), passes with the fix; also covers a diffuse write while unlit and re-enabling lighting. The existing `clone` case (diffuse colour + `disableLighting`) still passes.

Ran focused: `vitest --project compat materials.test.ts` (20 tests), eslint, prettier, `tsc` for the compat sources and tests.


bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260918.38/lab/index.html)

<sub>Build 20260918.38 - merge @ a99be6f5f2a1abeefcae85d4382934e149c72fed</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/750/)

<sub>Build 20260918.38 - merge @ a99be6f5f2a1abeefcae85d4382934e149c72fed</sub>

deltakosh (changes requested):
Hi Denis, thanks for the focused compatibility fix. One live-material regression still needs adjustment: `disableLighting` changes the Standard shader feature set, but this setter only marks the UBO dirty. After a material has rendered, toggling lighting off leaves the cached lit shader active while `_syncDiffuse()` uploads white, so scene lights incorrectly contribute to the result (for example, blue emissive plus a red light becomes magenta). Please rebuild/invalidate the material renderables when this mode changes so the shader variant and compensated diffuse update together. Thanks!

sdoleg2011:
Good catch, thank you — lighting on/off is a shader variant, so the UBO-only path was wrong for a material that has already rendered.

The `disableLighting` setter now goes through `_refreshInScene()` (`rebuildMaterial`) when the value actually changes, so the unlit variant and the compensated diffuse switch together; setting the same value again does not rebuild, and it stays a no-op until the scene has started.

Two regression cases in `material-texture-rebind.test.ts` (mocked Lite): toggling on a live material rebuilds exactly once per change and restores the app's diffuse when lighting is switched back on; nothing is rebuilt before the scene starts. The first one fails without the setter change.

Branch rebased onto current master. Re-ran focused: `material-texture-rebind.test.ts` + `materials.test.ts` (32 tests), eslint, prettier, `tsc` for the compat sources and tests.

deltakosh (changes requested):
Hi Denis, the live rebuild still misses Babylon.js's supported scene-less construction path. When 
ew StandardMaterial(name) is assigned to a mesh before startup, the mesh setter skips _adoptScene() because the scene is not started, and the deferred ddPrimitive() callback never adopts it. A later disableLighting toggle therefore leaves _scene undefined, so _refreshInScene() returns without rebuilding and the cached lit shader remains active. Please adopt the mesh scene during pre-start assignment or deferred registration, and cover scene-less construction followed by a post-start toggle. Thanks for tightening this last lifecycle case!

sdoleg2011:
Right — the scene-less path never reached `_adoptScene()`: before startup the `material` setter skips it, and the deferred add only called `_ensureRenderable()`.

The deferred add in `addPrimitive()` now adopts the mesh's scene for the final pre-start material, right before `_ensureRenderable()`. I put it there rather than in the pre-start setter so that a material which is assigned and then replaced before startup does not get registered with the scene. As a side effect the same path also fixes a scene-less material whose texture finishes loading after startup (it goes through the same `_refreshInScene()`).

New case in `material-texture-rebind.test.ts`: `new StandardMaterial(name)` with no scene, assigned to a `MeshBuilder.CreateBox` mesh before startup, scene goes live and flushes its deferred adds, then `disableLighting = true` — the material is registered with the scene and `rebuildMaterial` runs once with the scene's Lite handle. It fails without the change.

Since the mesh add path is shared I ran the whole compat project this time: 60 files / 714 tests pass; eslint, prettier, `tsc` for the compat sources and tests are clean. Branch rebased onto current master.

deltakosh (changes requested):
Thanks, the MeshBuilder path now adopts the scene correctly, but the same lifecycle blocker remains on supported deferred mesh paths that bypass `addPrimitive()`. `createDebugNavMesh(scene)` returns a mesh whose material can be replaced before startup, and `CSG.toMesh(name, material, scene)` accepts an explicit material; their deferred callbacks only call `_ensureRenderable()`/`addToScene()` and never `_adoptScene()`.

With a scene-less `StandardMaterial`, `getScene()` therefore stays undefined after startup and a later `disableLighting` toggle still produces no `rebuildMaterial()` call, leaving the cached lit shader active. Please make adoption apply to every supported registration path (or centralize it) and add regression coverage for these two bypasses.

Merci, la correction principale est bonne — il reste juste à couvrir ces chemins spécialisés.

sdoleg2011:
Thanks, David — you are right, and rather than patch two more call sites I centralized it (00ed4eda, rebased onto current master).

`attachMaterialToScene(material, scene)` in `materials.ts` is now the one place a compat material is made ready to render in a scene: adopt the scene, then `_ensureRenderable()`. I audited every spot where a compat material's Lite handle is put on a Lite mesh, and all five go through it:

- the live branch of the `material` setter (`meshes.ts`),
- the deferred add in `addPrimitive()` (`meshes.ts`) — MeshBuilder, `clone`, `MergeMeshes`, `LinesMesh`,
- `deferAddCsgMesh()` for `CSG.toMesh(name, material, scene)` (`csg.ts`),
- the source-material loop in `CSG2.toMesh(name, scene)` (`csg.ts`) — not deferred, but the same gap,
- `createDebugNavMesh(scene)` (`navigation.ts`).

The remaining `addToScene()` / `_deferAdd()` callers register lights, transform nodes, loaded containers, particle systems and sprites, none of which carry a caller-supplied compat material.

Regression coverage is a new GPU/wasm-free file, `material-scene-adoption.test.ts`, with one case per bypass (`CSG.toMesh`, `CSG2.toMesh`, `createDebugNavMesh`): a scene-less unlit `StandardMaterial`, scene not started, deferred adds flushed at start, then `disableLighting = true` → the material is registered with the scene, `getScene()` returns it, and `rebuildMaterial` runs once with the scene's Lite handle. The two cases you named fail on the previous revision. The `MeshBuilder` case from the last round stays in `material-texture-rebind.test.ts`.

Ran the whole compat project again since these are shared paths: 61 files / 717 tests pass; eslint, prettier and `tsc` for the compat sources and tests are clean.

deltakosh (changes requested):
The new helper fixes the covered registration callbacks, but scene adoption is still gated by `_hasStarted` in the mesh material setter, and several supported pre-start assignments have no later callback that adopts the effective material:

- assign a scene-less `StandardMaterial` to an imported mesh after `_bindLoadedScene()`;
- call `CSG.toMesh(name, null, scene)`, then assign the returned mesh’s material before startup (the deferred callback only sees its captured null argument);
- replace `scene.defaultMaterial` with a scene-less material, then set a primitive’s `material = null` (deferred registration reads `mesh.material`, not the effective default).

All three paths bind the Lite material for rendering but leave `getScene()` undefined, so a later `disableLighting` toggle or texture-ready callback still performs no rebuild and can leave the cached shader stale. Please decouple scene ownership from GPU finalization: adopt the effective render material whenever a mesh already has a scene, while keeping `_ensureRenderable()` gated to live/deferred registration, and cover these pre-start paths.

Merci — la centralisation va dans le bon sens, il reste ces affectations qui n’atteignent jamais les nouveaux callbacks.

sdoleg2011:
Thanks, David — agreed, tying ownership to the registration callbacks was the wrong cut. Reworked in 7aa00bbf (rebased onto current master): ownership and GPU finalization are now separate.

**Ownership** — `adoptMaterialScene()`. A mesh that has a scene adopts its *effective* render material (`value ?? scene.defaultMaterial`) in the `material` setter whether or not the scene has started, and the `scene.defaultMaterial` setter adopts a replacement. That covers the three paths you listed:

- a scene-less material given to an imported mesh after `_bindLoadedScene()` — there is no callback at all there, so assignment is the only place it can happen;
- `CSG.toMesh(name, null, scene)` followed by `result.material = mat` before startup;
- a scene-less `scene.defaultMaterial` rendered by a primitive whose `material` is `null`.

**Finalization** — `attachMaterialToScene()` (ownership + `_ensureRenderable()`), still gated to the live branch of the setter and to the deferred registration. That registration is now one function, `registerMeshAtStart()`, shared by the primitives (`addPrimitive` delegates to it), legacy CSG and `createDebugNavMesh`, and it finalizes and rebinds the *effective* material — `mesh.material ?? scene.defaultMaterial` as it is at startup — instead of only `mesh.material`. For legacy CSG this also removes the captured-argument problem at the root: `toMesh(name, material, scene)` now assigns the explicit material to the result (as Babylon.js does) and lets the shared registration pick up whatever the mesh's material is by then. `CSG2.toMesh` keeps its source materials, but a material given to the returned mesh before startup is finalized and bound as well.

One existing expectation changed on purpose: `material-texture-rebind.test.ts` had "only rebinds (no ensure/adopt) before the engine has started". It now asserts the split — adopted and rebound, not finalized.

Coverage in `material-scene-adoption.test.ts` (8 cases): the three paths above, the `scene.defaultMaterial` setter on its own, and the CSG / CSG2 / navigation builders, each ending with a post-start `disableLighting` toggle that must rebuild in the owning scene; where it applies they also check that `_ensureRenderable()` is *not* called before startup and *is* called with the engine at startup, and that the Lite mesh ends up with the right material handle. 7 of the 8 fail on the previous revision.

Whole compat project: 61 files / 722 tests pass; eslint, prettier, `tsc` for the compat sources and tests are clean.

deltakosh (changes requested):
The live rebuild now exposes an invalid Standard shader variant when `disableLighting` is enabled on a material with a ready `bumpTexture`. The unlit template removes the declaration of `normalW`, but the active normal-map fragment still emits `normalW = perturbNormal(...)`; WebGPU therefore rejects the rebuilt shader and the affected mesh stops rendering.

Please make bump-map composition valid for the unlit variant—either suppress the bump fragment/bindings when lighting is disabled or preserve the required normal declaration—and add a shader-level regression test for toggling this combination. The current tests mock `rebuildMaterial` and only exercise texture-free toggles, so they cannot catch WGSL validation failures.

Merci — l’adoption de scène est maintenant bien couverte ; il reste cette combinaison shader déclenchée par le rebuild live.

sdoleg2011:
Thanks, David — confirmed, and it is a little wider than the bump map. Fixed in e5f33283 (rebased onto current master).

**Cause.** The `AC` and `AT` fragment slots exist in both Standard variants. A bound normal map writes `normalW` in `AC`, a bound specular map writes `specularColor` in `AT`, and the unlit template dropped the declaration of both. So unlit + bump *and* unlit + specular map produced WGSL that assigns to an undeclared identifier. (The `AD`-slot fragments — ambient, reflection, shadows — are not affected: that slot only exists inside the lit lighting block, so their code is never emitted for the unlit variant.)

**Fix.** Of your two options I took "preserve the declaration": `standard-template.ts` now declares `normalW` and `specularColor` in the unlit variant too. My reasoning: it is one fix at the slot contract rather than one per extension, so any current or future fragment that writes those locals stays valid in both variants, and nothing about bindings or feature bits changes between the two states of a live toggle. Both locals are unused in the unlit colour path, so the rendered output is unchanged.

One follow-on simplification in the same commit: with `normalW` always declared, `composeStandardGeometryShader` no longer needs the unlit special case I added in #747, so normal attachments read `normalW` in both variants again. That also gives unlit materials bump-perturbed geometry normals. I updated the #747 test accordingly (it now asserts `normalW` is declared and used for both variants, which is the property that matters there).

**Shader-level regression** in `shader-integration.test.ts`: a real Standard material with a bump and a specular texture set through the public setters, `disableLighting` toggled `false → true → false`; for each state it composes exactly what the renderable composes (`_computeStandardMaterialFeatures` → registered extensions' fragments → `composeStandardShader`) and checks that every local a map fragment writes is declared before it is written. It fails on the previous revision with `normalW is declared: expected -1 …`.

Ran focused: `shader-integration` + `standard-geometry-features` + `geometry-renderer-task` (91 tests), `xr-teleport` (the other suite that uses unlit Standard materials, 25 tests), eslint, prettier, `tsc` for `packages/babylon-lite`. Local per-module accounting: `standard-template.js` −28 B on every Standard scene (two conditionals gone), and a further −41 B in `standard-geometry-output-shader.js` on the geometry-renderer scenes.

deltakosh (approved):
The unlit Standard shader now keeps the map-written locals declared across forward and geometry variants, and the focused regression coverage exercises both normal and specular maps. This resolves my blocker. Thanks!

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/750/)

<sub>Build 20260919.7 - merge @ 655d75ad61eb6e91bbf4cbc187796cdcf1e2920a</sub>

bjsplat:
## Bundle Size Changes

### Decreases

| Scene 43 — Parametric Proximity Path<br/>`scene43` | 56 KB | 57 KB | -1 KB |
| Scene 103 — Thin-Instance Raycast Picking<br/>`scene103` | 62 KB | 63 KB | -1 KB |
| Scene 113 — Picking Precision<br/>`scene113` | 60 KB | 61 KB | -1 KB |
| Scene 205 - LWR Facing Billboards<br/>`scene205` | 58 KB | 59 KB | -1 KB |
| Scene 312 — Mesh Blending Orthographic Screen Depth<br/>`scene312` | 111 KB | 112 KB | -1 KB |

### Ceiling headroom

<details>
<summary>Tightest scenes repo-wide — 1 of 106 under 1.0 KB</summary>

| Scene 4 — Shadows<br/>`scene4` ⬅ moved by this PR | 71.02 KB | 71.70 KB | 697 B |
| Scene 110 — RTT with material override<br/>`scene110` ⬅ moved by this PR | 48.68 KB | 49.80 KB | 1.1 KB |
| Scene 18 — PCF Shadows<br/>`scene18` ⬅ moved by this PR | 58.22 KB | 59.50 KB | 1.3 KB |
| Scene 16 — Thin Instances<br/>`scene16` ⬅ moved by this PR | 46.90 KB | 48.30 KB | 1.4 KB |
| Scene 173 - Navigation Dynamic Obstacles<br/>`scene173` ⬅ moved by this PR | 62.10 KB | 63.50 KB | 1.4 KB |
| Scene 267 — StandardMaterial Vertex Colors<br/>`scene267` ⬅ moved by this PR | 40.24 KB | 41.70 KB | 1.5 KB |
| Scene 43 — Parametric Proximity Path<br/>`scene43` ⬅ moved by this PR | 56.44 KB | 57.93 KB | 1.5 KB |
| Scene 209 — LWR Physics<br/>`scene209` ⬅ moved by this PR | 51.71 KB | 53.30 KB | 1.6 KB |
| Scene 15 — SpotLights + Ground<br/>`scene15` ⬅ moved by this PR | 44.60 KB | 46.20 KB | 1.6 KB |
| Scene 268 — Orthographic Camera<br/>`scene268` ⬅ moved by this PR | 43.00 KB | 44.60 KB | 1.6 KB |

</details>

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260919.7/lab/index.html)

<sub>Build 20260919.7 - merge @ 655d75ad61eb6e91bbf4cbc187796cdcf1e2920a</sub>

PR #754, opened 2026-09-18
fix(compat): apply PBR albedo colour, metallic and roughness exactly once

sdoleg2011:
## Problem

A factor-only `PBRMaterial` (colours, no maps) is backed by 1×1 solid textures, because Lite's PBR pipeline samples `baseColorTexture` and `ormTexture` unconditionally. `_ensureRenderable` baked `albedoColor` / `metallic` / `roughness` into those texels and neutralized the factors "so each contribution is applied exactly once".

That holds only until the next write. The setters keep writing the *factors* (`set albedoColor` → `_lite.baseColorFactor`, `set roughness` → `_lite.roughnessFactor`), so any assignment after the first build multiplies the already-baked texel: colour², roughness², metallic². The getters also read back the neutralized `1` instead of what the app set.

Setting the same look again is very common. In my kitchen configurator the room code re-applies `albedoColor` on the floor and wall materials on every room refresh, so on `lite-compat` 1.30.0-preview:

| surface | albedo | Babylon.js | Lite |
|---|---|---|---|
| ceiling | 1.00 | 173 | 172 |
| walls | 0.86 | 205 | 191 |
| floor | 0.68 | 219 | 184 |

(8-bit frame brightness, same scene and lights.) The darker the albedo, the bigger the gap — which is what a squared colour looks like — and the ratio was the same under the sun alone, the hemispheric light alone and the environment alone. With the fix below all three surfaces match Babylon.js within one level.

## Fix

Keep the backing textures neutral white and leave the values in the factors. A value is then applied once no matter when or how often it is set, and `metallic` / `roughness` getters keep returning the app's value. No behaviour change for a material that is configured once before its first build: texel × factor is the same product either way.

## Tests

`packages/babylon-lite-compat/tests/material-texture-rebind.test.ts` — new case: configure a factor-only material, build it, set the same `albedoColor` / `roughness` again, build again. On `master` it fails (the solid texel carries the baked colour while the factor carries it too); with the fix the backing texels are neutral, the factors hold `[0.68, 0.67, 0.62, 1]`, `0.9`, `0`, and the getters report them. The existing expectation that the ORM backing texture is created as `(1, 1, 1)` still holds.

Ran focused: `vitest --project compat material-texture-rebind.test.ts materials.test.ts` (29 tests), eslint, prettier, `tsc` for the compat sources and tests.


deltakosh (approved):
Hi Denis, this keeps the fallback textures neutral while preserving the caller-visible factors, and the focused regression covers the repeated-setter path well. Looks good, thanks!

bjsplat:
### Lab - Static Site

[Open deployed site](https://snapshots-cvgtc2eugrd3cgfd.z01.azurefd.net/lite/20260918.33/lab/index.html)

<sub>Build 20260918.33 - merge @ 59806e612da0279ec03b47e2223f1b102e4dfaeb</sub>

bjsplat:

[Open deployed site](https://liteplayground.babylonjs.com/pr/754/)

<sub>Build 20260918.33 - merge @ 59806e612da0279ec03b47e2223f1b102e4dfaeb</sub>

